PR #29, opened 2023-05-18
Use artifact name when copying a binary or a library in the dist task

ntrinquier:


kraken-std/src/kraken/std/dist.py
```diff
@@ -73,12 +73,7 @@ def execute(self) -> None:
         output_file.parent.mkdir(parents=True, exist_ok=True)
         with wopen_archive(output_file, archive_type) as archive:
             for resource in self.resources.get():
-                if resource.options.arcname is not None:
-                    arcname = resource.options.arcname
-                elif isinstance(resource, BinaryArtifact):
```
ntrinquier, line 78: this is where the bug was: at this point, `resource` may no longer be a `BinaryArtifact`, it is necessarily a `ConfiguredResource` (as the type of the property attests), which has no overlap with `BinaryArtifact`

kraken-std/src/kraken/std/dist.py
```diff
@@ -264,3 +253,29 @@ def dist(
...
+
+    for task, task_resources in resources.items():
+        for resource in task_resources:
+            task_options = dependencies_map[
+                next(iter(dependencies_set.partitions()[task]))
+            ]
+            options = IndividualDistOptions(**vars(task_options))
+
+            if options.arcname is None:
+                if isinstance(resource, (BinaryArtifact, LibraryArtifact)):
```
ntrinquier, line 273: the solution is to move the logic where we construct the `ConfiguredResource`s, to set the `arcname` in the options there

note that `LibraryArtifact`s now _also_ get their `arcname` changed, which is hopefully not controversial

kraken-std/src/kraken/std/dist.py
```diff
@@ -264,3 +254,21 @@ def dist(
...
+    dependencies_set: TaskSet,
+) -> list[ConfiguredResource]:
+    configured_resources = []
+
+    for task, task_resources in resources.items():
+        for resource in task_resources:
+            task_options = dependencies_map[next(iter(dependencies_set.partitions()[task]))]
+            options = IndividualDistOptions(**vars(task_options))
+
+            if options.arcname is None and isinstance(resource, (BinaryArtifact, LibraryArtifact)):
```
ntrinquier, line 271: the solution is to move the logic where we construct the `ConfiguredResource`s, to set the `arcname` in the `options` there

note that `LibraryArtifact`s now _also_ get their `arcname` changed, which is hopefully not controversial

PR #30, opened 2023-05-18
Update dependencies before attempting to package with helm

callumPearce:


kraken-std/src/kraken/std/helm/helmapi.py
```diff
@@ -23,7 +23,7 @@ def helm_package(
         raise ValueError("output_file and output_directory cannot both be set")
 
     with contextlib.ExitStack() as exit_stack:
-        command = ["helm", "package", str(chart_path)]
+        command = ["helm", "package", str(chart_path), "--dependency-update"]
```
ntrinquier: @callumPearce can we make this configurable with a keyword argument `dependency_update` in `helm_package`? (for which defaulting to `True` is reasonable IMO)

```suggestion
        command = ["helm", "package", str(chart_path)]
        if dependency_update:
            command.append("--dependency-update")
```

callumPearce: Yes, definitely!

callumPearce: https://github.com/kraken-build/kraken-build/pull/30/commits/8493560cd7bec86287cb1e82e0fc8dfaa6412015

kraken-std/src/kraken/std/helm/helmapi.py
```diff
@@ -16,6 +16,7 @@ def helm_package(
     output_directory: Path | None = None,
     app_version: str | None = None,
     version: str | None = None,
+    update_chart_dependencies: bool = True
```
ntrinquier:
```suggestion
    update_chart_dependencies: bool = True,
```

ntrinquier (changes requested):

ntrinquier (approved):

PR #32, opened 2023-05-24
cc/shorten stacktrace for known errors

TheMightyFid:
Reduces stack-trace for ValueErrors created in main.py, aswell as hiding stack-traces for script errors.


kraken-core/src/kraken/core/cli/main.py
```diff
@@ -549,22 +549,27 @@ def main_internal(prog: str, argv: list[str] | None, pdb_enabled: bool) -> NoRet
...
-                assert False, args.query_cmd
+            _context, graph = None, None
+            try:
+                _context, graph = _load_build_state(
+                    exit_stack=exit_stack,
+                    build_options=build_options,
+                    graph_options=graph_options,
+                )
+            except ValueError as err:
+                print(colored(("> "), "magenta") + colored(str(err), "yellow", attrs=["bold"]))
```
NiklasRosenstein: We should probably exit the process after printing this? Means we can get rid of the `if None in (_context, graph):` as well.

NiklasRosenstein: nit: print to stderr

TheMightyFid:
Update 26/05: After some trials with Python's exceptions + tracebacks, this should now significantly shorten these errors. @cdbrkfxrpt + @NiklasRosenstein, this should be ready for review.

cdbrkfxrpt (approved):

kraken-common/src/kraken/common/_runner.py
```diff
@@ -109,7 +109,10 @@ def execute_script(self, script: Path, scope: Dict[str, Any]) -> None:
         module = types.ModuleType(str(script.parent))
         module.__file__ = str(script)
 
-        code = compile(script.read_text(), script, "exec")
+        try:
+            code = compile(script.read_text(), script, "exec")
+        except BaseException as err:
+            raise ValueError("Script Failed to compile") from err
```
ntrinquier:
```suggestion
            raise ValueError("Script failed to compile the build script") from err
```

kraken-core/src/kraken/core/cli/main.py
```diff
@@ -628,22 +628,34 @@ def main_internal(prog: str, argv: list[str] | None, pdb_enabled: bool) -> NoRet
...
-            else:
-                assert False, args.query_cmd
+            optional_graph: Optional[TaskGraph] = None
+            try:
+                _, optional_graph = _load_build_state(
+                    exit_stack=exit_stack,
+                    build_options=build_options,
+                    graph_options=graph_options,
+                )
+            except ValueError as err:
```
NiklasRosenstein, line 622: If the intention of reraising as a `ValueError` in the above change is to catch it here, I would suggest to introduce some kind of dedicated exception type. A `ValueError` is normally used to indicate when an invalid value was passed (e.g.  `int("garble")`).

kraken-core/src/kraken/core/cli/main.py
```diff
@@ -628,22 +628,34 @@ def main_internal(prog: str, argv: list[str] | None, pdb_enabled: bool) -> NoRet
...
+                print(
+                    colored(("> "), "magenta")
+                    + colored("Kraken has detected an error: " + str(err), "yellow", attrs=["bold"])
+                )
+                if err.__cause__ is not None:
+                    sys.excepthook(type(err.__cause__), err.__cause__, err.__cause__.__traceback__)
+
+            if optional_graph is not None:
+                graph: TaskGraph = optional_graph
```
NiklasRosenstein, line 631: After the `if optional_graph is not None:` check, Mypy understands that `optional_graph` is not None. So there's not really a need to re-assign here.

kraken-core/src/kraken/core/cli/main.py
```diff
@@ -628,22 +628,34 @@ def main_internal(prog: str, argv: list[str] | None, pdb_enabled: bool) -> NoRet
...
+                    build_options=build_options,
+                    graph_options=graph_options,
+                )
+            except ValueError as err:
+                print(
+                    colored(("> "), "magenta")
+                    + colored("Kraken has detected an error: " + str(err), "yellow", attrs=["bold"])
+                )
+                if err.__cause__ is not None:
+                    sys.excepthook(type(err.__cause__), err.__cause__, err.__cause__.__traceback__)
```
NiklasRosenstein, line 628: If there's an error, I don't see any reason for Kraken to try and continue? I suppose it'll never really, because there's only one statement that assigns to `optional_graph` so if that fails there's no graph. Let's just return early then?

kraken-common/src/kraken/common/_runner.py
```diff
@@ -109,7 +109,10 @@ def execute_script(self, script: Path, scope: Dict[str, Any]) -> None:
         module = types.ModuleType(str(script.parent))
         module.__file__ = str(script)
 
-        code = compile(script.read_text(), script, "exec")
+        try:
+            code = compile(script.read_text(), script, "exec")
+        except BaseException as err:
+            raise ValueError("Script failed to compile the build script") from err
```
NiklasRosenstein, line 115: This doesn't feel quite right. For one, because it catches any and all exceptions when really we probably just want to catch `SyntaxError`s. And what benefit does the reraise as a `ValueError` give us?

kraken-core/src/kraken/core/cli/main.py
NiklasRosenstein: Just recently I added the `on_exception()` function above `main()`, I think that handles (most of?) the cases you were looking to catch here.

kraken-core/src/kraken/core/cli/main.py
TheMightyFid: @NiklasRosenstein I ran with the assumption that there *could* be additional code to be run (`closing log files`, etc) and relied upon the bounding code to handle `null`s. Can add an early exit here though if this isn't a concern

kraken-common/src/kraken/common/_runner.py
TheMightyFid: Agree with catching all errors being a bit of a nuclear option here, but assumed as it was only 'compiling' and not running the python code, that this would be OK. Catching `Syntax Error` is a good suggestion, but perhaps also `Assertion`, `Import`, `ModuleNotFound`, `Indentation`, `Type`, `Unicode` and `ZeroDivision` errors would also be good?
The re-raise gives us the cleaner error message to handle later and give to the end-user. It'll get unrolled to give the original error if we want to display it;

kraken-core/src/kraken/core/cli/main.py
TheMightyFid: Ah - I remember the issue now (it's been a few weeks). The `graph` object stops `mypy` from complaining about the type differences between the input type to `ls(graph)` => `TaskGraph`, and the actual type of `optional_graph` => `Optional[TaskGraph]`. If there's a better way to unroll this here, I'd give that a go.

kraken-core/src/kraken/core/cli/main.py
TheMightyFid: Cleaned and removed in latest commit -

TheMightyFid: Cleaned and removed in latest commit -

kraken-core/src/kraken/core/cli/main.py
TheMightyFid: Cleaned and removed in latest commit -

kraken-core/src/kraken/core/cli/main.py
TheMightyFid: Now Removed

kraken-core/src/kraken/core/cli/main.py
TheMightyFid: Removed

TheMightyFid:
I'm closing this as it looks like kraken has moved on from the outdated code in this PR. The now only appears to be in _runner.py

PR #37, opened 2023-06-08
Disable the nightly formatter by default

j-baker:
Hi! I saw the default nightly formatting feature merge earlier and I realised that I was kind of hosed. Why? I don't use Rustup, and this feature relies on Rustup, so my kraken fmt doesn't work anymore.

I'm also a bit concerned that the behaviour as implemented default moves every dev team onto nightly Rust, which I don't think we should be encouraging without careful thought and some tooling.

I'd therefore like us to take a quick step back by making it be an opt-in and not an opt-out feature so as to give more time for the above attributes to be addressed. Sound good?

It feels like the correct default behaviour would be to add a 'nightly' flag that runs _all_ commands with nightly rust, rather than just fmt, and so one can opt a project into nightly rust and not just the formatter.


j-baker:
I believe the glitch is here:

https://github.com/kraken-build/kraken-build/blob/develop/kraken-std/src/kraken/std/cargo/tasks/cargo_fmt_task.py#L42

where it fails if nightly is not the toolchain in use.

PR #38, opened 2023-06-08
Only check the active toolchain if rustup is installed

ntrinquier:


cdbrkfxrpt (approved):

j-baker:
I don't think this is quite right. I expect the right solution for this problem is to do something like run `rustc --version` and check that the version string is a nightly release, if you want to ensure that everything runs with nightly.

It also still checks that all compilation occurs with nightly if one _does_ have rustup, I think? Whereas I think the `+nightly` will dynamically enable it for the command, so we only really want to check if nightly is _installed_, not that it's active.

If one were checkign with `rustup`, I'd expect that the right way would be to list the installed toolchains and check that nightly is in the list?

ntrinquier:
> I don't think this is quite right. I expect the right solution for this problem is to do something like run `rustc --version` and check that the version string is a nightly release?
> 
> It also still checks that all compilation occurs with nightly, I think?

This only affects `cargo fmt`, not the compilation (and now that we made it opt-out, only the user specified `nightly=True`)

j-baker:
Won't `rustup show active-toolchain` return nightly if rustc is going to compile with nightly? Or am I misunderstanding what that command returns?

ntrinquier:
> I think the +nightly will dynamically enable it for the command, so we only really want to check if nightly is installed, not that it's active.

yeah that's fair

cdbrkfxrpt (approved):

PR #43, opened 2023-06-19
Support running tests in parallel

ntrinquier:


ntrinquier:
Actually we can defer to the users of `pytest()` to decide if they want to run tests in parallel, which they can do with `PYTEST_FLAGS="--numprocesses auto"`

PR #46, opened 2023-06-20
implement poetry index priority in pyproject

samrogerson:


kraken-std/.kraken.lock
```diff
@@ -5,7 +5,7 @@ requirements = [
 pythonpath = [
     "build-support",
 ]
-interpreter_constraint = ">=3.10,<3.11"
+interpreter_constraint = ">=3.7"
```
NiklasRosenstein: Can you upgrade your Kraken-wrapper version? :)

Unfortunately the interpreter constraint is not currently informed by the dependencies (as `kraken-core` now requires 3.10 since 0.25.0), but it's put there by `krakenw` if it's not specified in the `buildscript()` call.

kraken-std/src/kraken/std/python/pyproject.py
```diff
@@ -161,7 +161,7 @@ def get_name(self) -> str:
     def get_version(self) -> str | None:
         return cast(str, self._pyproj["project"].get("requires-python", None))
 
-    def upsert_source(self, source_name: str, url: str, _default: bool = False, _secondary: bool = False) -> None:
+    def upsert_source(self, source_name: str, url: str, priority: str = "") -> None:
```
NiklasRosenstein: Can we use `IndexPriority` here as well?

kraken-std/src/kraken/std/util/check_file_exists_and_is_committed_task.py
```diff
@@ -18,7 +18,7 @@ def to_description(self, file_path: Path) -> str:
         if self == CheckFileExistsAndIsCommittedError.IS_NOT_COMMITTED:
             return f"'{file_path}' exists but is not committed"
 
-        return "IMPOSSIBLE STATE"
+        return "IMPOSSIBLE STATE"  # type: ignore[unreachable]
```
NiklasRosenstein: Tbh. I don't know why we had it like this here in the first place, imo this should just `assert False`.

```suggestion
        assert False, "unreachable"
```

(Not sure if the `type: ignore[unreachable]` comment needs to stay)

kraken-std/src/kraken/std/util/check_valid_readme_exists_task.py
```diff
@@ -47,7 +47,7 @@ def to_description(self) -> str:
             )
 
         # This should be an impossible state
-        return "Unknown `readme` error."
+        return "Unknown `readme` error."  # type: ignore[unreachable]
```
NiklasRosenstein: Same as above.

kraken-std/src/kraken/std/util/check_file_exists_and_is_committed_task.py
samrogerson: Agreed; can do in a different PR but mypy throws when this isn't here now

kraken-std/src/kraken/std/python/buildsystem/slap.py
```diff
@@ -116,7 +117,7 @@ def install(self, settings: PythonSettings) -> None:
                 if index.credentials:
                     spec += f",username={quote(index.credentials[0])},password={quote(index.credentials[1])}"
                 safe_spec = spec.replace(quote(index.credentials[1]), "[MASKED]") if index.credentials else spec
-                option = "--index" if index.default else "--extra-index"
+                option = "--index" if index.priority == IndexPriority.default else "--extra-index"
```
samrogerson: enum comparison :/ whoops

samrogerson: we should probably lint for this :|

kraken-std/src/kraken/std/python/pyproject.py
samrogerson: it creates a circular import

samrogerson: pulled out

kraken-std/src/kraken/std/python/buildsystem/slap.py
samrogerson: but, fixed now

kraken-std/src/kraken/std/util/check_file_exists_and_is_committed_task.py
NiklasRosenstein: 👍 We pick up the latest Mypy version always so whatever they add in terms of new checks, it may require some code adjustments every now and then.

NiklasRosenstein (approved):

PR #48, opened 2023-06-21
pytest coverage

bpoumarede:


kraken-std/pyproject.toml
```diff
@@ -39,6 +39,7 @@ isort = "*"
 mypy = "*"
 pycln = "*"
 pytest = "*"
+pytest-cov = "*"
```
NiklasRosenstein: We only need this here if we actually make use of it in the build script for this project. Would you mind updating the `.kraken.py` script as well and integrate the Coverage report into CI?

kraken-std/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -42,6 +58,13 @@ def get_execute_command(self) -> list[str] | TaskStatus:
         ]
         command += flatten(["--ignore", str(self.project.directory / path)] for path in self.ignore_dirs.get())
         command += ["--log-cli-level", "INFO"]
+        if self.coverage.get():
+            coverage_file = f"coverage{get_coverage_path(self.coverage.get())}"
+            command += [
+                "--cov-report",
+                f"{self.coverage.get()}:{str(self.project.build_directory / coverage_file)}",
+                f"--cov={str(self.project.directory / tests_dir)}",
```
NiklasRosenstein: The `--cov` option needs to point to the source directory

kraken-std/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -7,10 +7,25 @@
...
 
-# TODO (@NiklasRosenstein): Pytest coverage support
+CoverageFormat = Literal["xml", "html", "json", "lcov", "annotate", None]
+
+
+def get_coverage_path(format: CoverageFormat) -> str:
+    if format == "xml":
+        return ".xml"
+    elif format == "html":
+        return "_html"
```
NiklasRosenstein: Why is it `_html` and not `.html`?

kraken-std/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -7,10 +7,25 @@
...
 from kraken.core import Project, Property, TaskStatus
+from typing_extensions import Literal
 
 from .base_task import EnvironmentAwareDispatchTask
 
-# TODO (@NiklasRosenstein): Pytest coverage support
+CoverageFormat = Literal["xml", "html", "json", "lcov", "annotate", None]
+
+
+def get_coverage_path(format: CoverageFormat) -> str:
```
NiklasRosenstein: A better name for this function may be `get_coverage_suffix()`? 

An enum may be a better place for the coverage formats and this method to live in.

kraken-std/src/kraken/std/python/tasks/pytest_task.py
bpoumarede: Because for html and annotate pytest coverage creates a folder instead of a file.

kraken-std/src/kraken/std/python/tasks/pytest_task.py
bpoumarede: I first created an enum but Nick thought it was a better idea to use a Literal. I will speak with him about it.

kraken-std/pyproject.toml
bpoumarede: Yes I will add that

kraken-std/src/kraken/std/python/tasks/pytest_task.py
bpoumarede: You are right, I will change that

kraken-std/src/tests/integration/python/test_python.py
```diff
@@ -176,3 +176,30 @@ def test__python_pyproject_reads_correct_data(
...
+    local_build_system = python.buildsystem.detect_build_system(tempdir)
+    assert local_build_system is not None
+    assert local_build_system.get_pyproject_reader(pyproject) is not None
+    assert local_build_system.get_pyproject_reader(pyproject).get_name() == "slap-project"
+
+    python.settings.python_settings(project=kraken_project, build_system=local_build_system)
+    python.pytest(project=kraken_project, coverage=python.CoverageFormat.XML)
+    python.install(project=kraken_project)
+    python.build(project=kraken_project)
+    kraken_ctx.execute([":build", ":test"])
```
NiklasRosenstein, line 205: The test looks good! Only missing a check to see if the coverage file exists after execute().

bpoumarede: Yes ! I realised it. I just added it now.

NiklasRosenstein: We may want to consider also checking the contents of the coverage report, to ensure we've passed the right directory etc. But gonna merge for now. :)

NiklasRosenstein (approved):
We may want to consider also testing that the coverage report contains what it should contain, but I'm happy to merge it in its current state to unblock downstream work.

kraken-std/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -42,6 +56,13 @@ def get_execute_command(self) -> list[str] | TaskStatus:
         ]
         command += flatten(["--ignore", str(self.project.directory / path)] for path in self.ignore_dirs.get())
         command += ["--log-cli-level", "INFO"]
+        if self.coverage.is_filled():
+            coverage_file = f"coverage{self.coverage.get().get_suffix()}"
+            command += [
+                "--cov-report",
+                f"{self.coverage.get().get_format()}:{str(self.project.build_directory / coverage_file)}",
+                f"--cov={str(self.project.directory / self.settings.source_directory)}",
+            ]
```
antoinebrl, line 65: Should there be a check to see if `pytest-cov` is installed before adding these parameters?

PR #52, opened 2023-06-23
Bump Cargo.toml version before building

daladim:
Before this commit, Cargo.toml was only bumped before _publishing_, but the binary would already have been built and cached, and any use of e.g. `env!("CARGO_PKG_VERSION")` would return unexpected values


kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -252,7 +252,7 @@ def cargo_bump_version(
     version: str,
     revert: bool = True,
     name: str = "cargoBumpVersion",
-    group: str | None = CARGO_PUBLISH_SUPPORT_GROUP_NAME,
```
ntrinquier, line 255: I think the task still needs to belong to `CARGO_PUBLISH_SUPPORT_GROUP_NAME` (you can add `task.depends_on(f":{CARGO_BUILD_SUPPORT_GROUP_NAME}?")` after the task has been created)

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -270,14 +269,14 @@ def cargo_bump_version(
     task = project.do(
         name,
         CargoBumpVersionTask,
-        group=group,
+        group=CARGO_BUILD_SUPPORT_GROUP_NAME,
```
ntrinquier, line 277: could you also add a comment explaining why `CARGO_BUILD_SUPPORT_GROUP_NAME` is needed? the explaination in your PR description about `env!("CARGO_PKG_VERSION")` would suffice

ntrinquier (approved):

PR #53, opened 2023-06-23
Refactor internal `Pyproject` API

NiklasRosenstein:
* The `Pyproject` now only represents the raw TOML data
* The `PyprojectHandler` replaces `SpecializedPyproject` and serves as a base class that can deal with basic PEP 621 project metadata. The tool-specific implementations have moved into ther corresponding module under `kraken.std.python.buildsystem`.
* The `PyprojectHandler` API for dealing with package indexes now uses strong types and the `PackageIndex` abstraction dataclass instead of passing around untyped dicts. Instead of delete/upsert, we now have getter and setter methods for the package index list in a project.


kraken-std/src/kraken/std/python/pyproject.py
```diff
@@ -1,234 +1,151 @@
...
+class PackageIndex:
+    """
+    Represents a Python package index. Some tool-specific representations of package indices may not support
+    all fields, in which case the implementation should trigger a warning or error if the alternative behaviour
+    conflicts with the field value.
+    """
 
-        # TODO(@niklas.rosenstein): Support Poetry dependency groups
-        #       https://python-poetry.org/docs/master/managing-dependencies/#dependency-groups
+    Priority: ClassVar[TypeAlias] = _PackageIndexPriority
```
samrogerson, line 38: why do the indirect `TypeAlias` rather than just directly use `PackageIndexPriority`?

NiklasRosenstein: Simplifying the API; i.e. you only need to import `PackageIndex` and now have access to the `.Priority` type. I would usually just nest the enum definition into the class but that sometimes causes issues with deserialization.

PR #69, opened 2023-07-08
Set ssl-insecure on mitm-proxy

tetigi:
Reasoning - this proxy is _only_ used for Cargo registry handling, and thus doesn't really need to be validating the certs.

This 'fixes' issues where the user's certs are self-signed. **There is probably a better way to do this.**


ntrinquier:
@tetigi do we still need this? For self-signed certificates, perhaps we should expect users to add their CA certs on their system?

PR #73, opened 2023-07-11
nix flake redux

j-baker:
some notes - adds a poetry lock to kraken-wrapper - nix needs to know exactly what versions of deps to use and can read a poetry lockfile, poetry lock seems fairly standard.


kraken-wrapper/pyproject.toml
```diff
@@ -20,7 +20,7 @@ Homepage = "https://kraken-build.github.io/kraken-build/"
 Repository = "https://github.com/kraken-build/kraken-build/"
 
 [tool.poetry.dependencies]
-python = ">=3.7"
+python = ">=3.7,<3.11"
```
NiklasRosenstein:
```suggestion
python = ">=3.7"
```

flake.nix
```diff
@@ -0,0 +1,31 @@
+{
```
NiklasRosenstein, line 1: Should this file not live in the `kraken-wrapper/` directory?

kraken-wrapper/poetry.lock
```diff
@@ -0,0 +1,900 @@
+# This file is automatically @generated by Poetry 1.4.2 and should not be changed by hand.
```
NiklasRosenstein, line 1: How are we making sure this file stays up to date when our everyday workflow does not include Poetry?

flake.nix
j-baker: It's easier for others to depend on it if it lives in the root - if it's in the root you can just write github:kraken-build/kraken-build to depend on it, if it's in a subdir you have to push that path, which also means more refactoring. If we add more nixed tests for kraken, this'll be a good place to put them, too.

kraken-wrapper/pyproject.toml
j-baker: poetry doesn't like this. This is because:
1. there are dependencies that don't support python 4.
2. there are dependencies that don't support python 3.12
3. I believe that kraken breaks if you're on 3.11 anyway? although the wrapper might be fine.

kraken-wrapper/poetry.lock
j-baker: CI runs `poetry lock --check`

kraken-wrapper/pyproject.toml
NiklasRosenstein: Python 4?? xD If Poetry has an issue with `>=`, we can make it `^3.7` instead, or add `<4.0`.

The wrapper is indeed fine on 3.11 and just yesterday I made the PR because someone was facing the issue about not being able to install it into 3.11. The wrapper has no issue with 3.11 so I considered the upper limit a bug.

If there are dependencies that don't support 3.12 we can still add back `<3.12` until they support it.

flake.nix
NiklasRosenstein: Hm, the main thing that worries me is that the repository contains more than the wrapper, but the Nix flake only gives you the wrapper. 

A note on the side, can the flake be referenced by a Git tag instead? I'd like to avoid stuff referencing `HEAD` and us having to forever be backwards compatible or support the flake in that location at all.

kraken-wrapper/pyproject.toml
j-baker: ok i'll make it <3.12. It is indeed an issue to just do ^3.7, because poetry picks dependencies that are valid for all versions in the range (I just figured out why they do this).

I think this change would actually be beneficial for all kraken users if applied in kraken-std etc - mostly i'd be worried about different python versions producing differnet kraken lockfiles

flake.nix
j-baker: I think it's ok for the flake to only give you the wrapper fwiw. I think later we'd probably add some 'checks' to the nix flake, but I don't think there's any inconsistency here wrt other nix things. This is just the flexible choice, and after all, the flake is kind of the entrypoint of Kraken.

It's definitely possible to refer to flakes by git tags. Most people select a branch because the nix flakes have a lockfile so it only updates on demand. The risk in doing this is obviously that any kraken commit is now something someone might depend on, rather than only releases, but the workflow is very slick on the nix side and renovate can handle the lock updating for you.

kraken-wrapper/pyproject.toml
NiklasRosenstein: > I think this change would actually be beneficial for all kraken users if applied in kraken-std etc - mostly i'd be worried about different python versions producing differnet kraken lockfiles

That would be beneficial, if it would impact `.kraken.lock` file generation at all. Currently we just put into the lock file whatever Pip installs, and that's what the current Python version generates. So far we haven't run into issues connected to this, but admittedly that does involve a bit of luck.

flake.nix
NiklasRosenstein: Ok, I feel more comfortable knowing that there's lock files involved. What I want to avoid is that we do end up deciding to move the flake into a sub directory, or changing what it produces etc. (i.e. making a backwards incompatible change with how it is consumed) and breaking the world because everything depends on "latest".

j-baker: yes - that doesn't happen with Nix. You have to update the lockfiles eventually, though!

NiklasRosenstein (approved):

PR #77, opened 2023-07-13
Allow passing of enabled features to cargo build

morosanmihail:
This allows cargo_build to take a list of strings (features) to pass onto the build task and enable them during compilation.

Despite some examples online recommending to wrap all comma-separated features in a pair of quotes ("abc, bcd"), the official documentation does not mention this as a requirement. To be revisited if problematic.


kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -322,6 +324,9 @@ def cargo_build(
         additional_args.append(crate)
     if mode == "release":
         additional_args.append("--release")
+    if features:
+        additional_args.append("--features")
+        additional_args.append(",".join(features))
```
ntrinquier, line 329:
```suggestion
        additional_args.extend(features)
```

so that the final command looks like `["--features", "feature1", "feature2"]` rather than `["--features feature1 feature2"]`

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -308,7 +309,8 @@ def cargo_build(
         variables in :attr:`CargoProject.build_env`.
     :param exclude: List of workspace crates to exclude from the build.
     :param name: The name of the task. If not specified, defaults to `:cargoBuild{mode.capitalised()}`.
-    :param version: Bump the Cargo.toml version temporarily while building to the given version."""
+    :param version: Bump the Cargo.toml version temporarily while building to the given version.
```
ntrinquier: looks like `version` no longer is a parameter, do you mind cleaning it up?

ntrinquier (changes requested):

kraken-std/src/kraken/std/cargo/__init__.py
morosanmihail, line 337: According to https://doc.rust-lang.org/cargo/reference/features.html , it needs to look like `--features foo-feat,bar-feat`, so the final command needs to look like `["--features", "feature1,feature2"]`.

ntrinquier: from the [docs](https://docs.python.org/3/library/subprocess.html#frequently-used-arguments)

> Providing a sequence of arguments is generally preferred, as it allows the module to take care of any required escaping and quoting of arguments (e.g. to permit spaces in file names).

ntrinquier: ah you beat me to it, I deleted my comment but too late; you're right, this is actually what we want here :)

do you mind leaving a comment which explains that?

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -322,6 +323,11 @@ def cargo_build(
         additional_args.append(crate)
     if mode == "release":
         additional_args.append("--release")
+    if features:
+        additional_args.append("--features")
+        # `cargo build`` expects features to be comma separated, in one string.
```
ntrinquier:
```suggestion
        # `cargo build` expects features to be comma separated, in one string.
```

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -322,6 +323,11 @@ def cargo_build(
         additional_args.append(crate)
     if mode == "release":
         additional_args.append("--release")
+    if features:
```
j-baker, line 333: out of interest, why are you seeking to enable these features in Kraken, rather than making them be default features in the root project. Basically, what's the use case for this functionality?

ntrinquier: @morosanmihail, I am keen to understand the use case as well :)

ntrinquier (approved):

kraken-std/src/kraken/std/cargo/__init__.py
ntrinquier: Capturing for posterity: this would be useful to enable platform-specific features ([a workaround](https://github.com/rust-lang/cargo/issues/1197#issuecomment-1641086954) exists, but it's not great), as well as to have multiple cargo build tasks to produce variants of a binary, with different features combinations.

PR #78, opened 2023-07-13
fix the poetry lock

j-baker:


ntrinquier (approved):

ntrinquier:
Done already. Having to do this after each release is annoying though, we should get rid of it :)

PR #84, opened 2023-07-24
Allows to use Maturin with PDM

Tpt:


kraken-std/src/kraken/std/python/buildsystem/maturin.py
```diff
@@ -196,14 +148,89 @@ def build(self, output_directory: Path, as_version: str | None = None) -> list[P
...
             pyproject.set_version(old_poetry_version)
             pyproject.raw.save()
 
         return dst_files
 
 
-class MaturinManagedEnvironment(PoetryManagedEnvironment):
+class MaturinPoetryPyprojectHandler(PoetryPyprojectHandler):
+    def set_version(self, version: str | None) -> None:
+        PyprojectHandler.set_version(self, version)
```
NiklasRosenstein, line 145: Shouldn't be necessary as it's called by the Poetry implementation is necessary?

kraken-std/src/kraken/std/python/buildsystem/maturin.py
```diff
@@ -212,3 +239,56 @@ def install(self, settings: PythonSettings) -> None:
 
     def always_install(self) -> bool:
         return True
+
+
+class MaturinPdmPythonBuildSystem(PDMPythonBuildSystem):
+    """A maturin-backed version of the PDM build system, that invokes the maturin build-backend.
+    Can be enabled by adding the following to the local pyproject.yaml:
+    ```toml
+    ~[tool.pdm.dev-dependencies]
```
NiklasRosenstein: Tilde intentional?

kraken-std/src/kraken/std/python/buildsystem/maturin.py
```diff
@@ -78,86 +52,51 @@ class MaturinZigTarget:
...
-        handler = self.get_pyproject_reader(pyproject)
-        handler.synchronize_project_section_to_poetry_state()
-
-    def build(self, output_directory: Path, as_version: str | None = None) -> list[Path]:
+    def build(
+        self,
+        output_directory: Path,
+        as_version: str | None,
+        get_pyproject_reader: Callable[[Pyproject], PyprojectHandler],
+        project_directory: Path,
```
ntrinquier: I know they are only used in `build`, but would it make sense to store `get_pyproject_reader` and `project_directory` (potentially `output_directory` too) in the constructor to simplify `build`'s signature?

kraken-std/src/kraken/std/python/buildsystem/maturin.py
Tpt: Sadly the Poetry implementation is not calling `super()`. So, it is not changing the version in the `[project]` section but only in the `[tool.poetry]` section.

kraken-std/src/kraken/std/python/buildsystem/maturin.py
Tpt: Indeed not. Fixed. Thank you!

ntrinquier (approved):

PR #87, opened 2023-07-25
Add Cargo Task to generate Deb packages

morosanmihail:
This creates a Kraken Task that allows users to generate Debian packages from a Cargo project. 

These packages can then be published as needed, most likely to apt repositories.


kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -405,16 +422,31 @@ def cargo_check_toolchain_version(
...
     """Creates a task that installs a given target for Cargo"""
 
     project = project or Project.current()
     task = project.task(f"rustupTargetAdd/{target}", RustupTargetAddTask, group=group)
     task.target = target
     return task
+
+
+def cargo_generate_deb_package(*, package_name: str):
+    project = project or Project.current()
```
ntrinquier: `project` is not defined here, you need:

```suggestion
def cargo_generate_deb_package(*, project: Project | None = None, package_name: str):
    project = project or Project.current()
```

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -405,16 +422,31 @@ def cargo_check_toolchain_version(
...
+    task = project.task(
+        f"cargoGenerateDeb",
```
ntrinquier: no need for a f-string:

```suggestion
        "cargoGenerateDeb",
```

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -405,16 +422,31 @@ def cargo_check_toolchain_version(
...
+        CargoGenerateDebPackage,
+        group=CARGO_PUBLISH_SUPPORT_GROUP_NAME,
```
ntrinquier: Does it actually need to be in this group?

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -405,16 +422,31 @@ def cargo_check_toolchain_version(
...
+    )
+    task.package_name = package_name
```
ntrinquier: If you call `project.do` you can actually pass the properties directly, something like:

```
project.do(
  "cargoGenerateDeb",
  ...
  package_name=package_name,
)

File: kraken-std/src/kraken/std/cargo/__init__.py (hunk 3 above)
Comment (morosanmihail) [reply kraken-std/src/kraken/std/cargo/__init__.py]: It *has* to happen after build. I can easily have this be dependent on the build tasks, but I was also told it being in the publish group solves that as well. What's preferred?

Comment (ntrinquier) [reply kraken-std/src/kraken/std/cargo/__init__.py]: Yep I'd put it in the `publish` group which has a dependency on the `build` group, so it will give you what you want. The publish _support_ group is a bit special, it's meant to be a dependency for the entire `publish` group :)

Comment (morosanmihail) [reply kraken-std/src/kraken/std/cargo/__init__.py]: Ah, got it! I thought that one was the `publish` group.

File: kraken-std/src/kraken/std/cargo/__init__.py [hunk 5]
```diff
@@ -405,16 +422,32 @@ def cargo_check_toolchain_version(
...
+
+def cargo_generate_deb_package(*, project: Project | None = None, package_name: str):
+    project = project or Project.current()
+    task = project.do(
+        "cargoGenerateDeb",
+        CargoGenerateDebPackage,
+        package_name=package_name,
+        group="publish",
+    )
+    task.depends_on(":cargoBuildRelease")
```
ntrinquier: It is a bit annoying because the task may not always be called `cargoBuildRelease`, it can also be `cargoBuildDebug`: https://github.com/kraken-build/kraken-build/blob/3011038a7b8c671c606c7c9f97e6b71771914a1c/kraken-std/src/kraken/std/cargo/__init__.py#L291-L292

Perhaps the cleaner solution here is to have `CargoGenerateDebPackage` accept a list of `package_name`s which can be `list[str] | list[Property[BinaryArtifact]]`.

Then, when calling `cargo_generate_deb_package`, you'd pass it the path to the binaries or libs coming from the build task:

https://github.com/kraken-build/kraken-build/blob/3011038a7b8c671c606c7c9f97e6b71771914a1c/kraken-std/src/kraken/std/cargo/tasks/cargo_build_task.py#L45-L49

Because if a task has a property which is another task's output property, kraken automatically creates a dependency between the two.

morosanmihail: I think this should be fine, as we only want to package release binaries for now. It could definitely be improved in the longer run.

ntrinquier: Still, we cannot leave it as is because sometimes the build task will not exist: for example, it might be `cargoBuildRelease` on develop, and `cargoBuildDebug` on a feature branch, in which case this will fail.

If we don't want to automatically generate the dependency via `Property`s, then we should defer to the caller (who will know better) the responsibility of adding the dependency

morosanmihail: Wouldn't we want this to fail if someone has defined it, but there's no build? The only way to use this is manual, for now.

ntrinquier: Unfortunately, it is not _the task_ that would fail, but the generation of the task graph. Meaning that you would not be able to run _any_ kraken task, as building the dependency graph would fail with an error (something like `No task matched the selector ':**:cargoBuildRelease'`)

morosanmihail: Just checked and I think I can safely remove the dependency on any `cargoBuild*`. `cargo deb` seems to build an optimised binary even if `cargo build` wasn't run first, so it will work in any situation.

And it being part of the `publish` group should limit repeating work.

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -423,3 +434,13 @@ def rustup_target_add(target: str, *, group: str | None = None, project: Project
...
     return task
+
+
+def cargo_generate_deb_package(*, project: Project | None = None, package_name: str) -> CargoGenerateDebPackage:
+    project = project or Project.current()
+    return project.do(
+        "cargoGenerateDeb",
+        CargoGenerateDebPackage,
+        package_name=package_name,
+        group="publish",
```
ntrinquier: I think only the task that actually publishes would be part of the group? 
```suggestion
```

kraken-std/src/kraken/std/cargo/tasks/cargo_generate_deb.py
```diff
@@ -0,0 +1,43 @@
...
+from kraken.std.descriptors.resource import LibraryArtifact
+
+
+@dataclass
+class CargoDebianArtifact(LibraryArtifact):
+    pass
+
+
+class CargoGenerateDebPackage(Task):
+    package_name: Property[str] = Property.default("CargoDebian")
```
ntrinquier: Why is this the default value? Don't we always want to force the caller to pass a package name (in which case there should not be a default)? 🤔

kraken-std/src/kraken/std/cargo/tasks/cargo_generate_deb.py
```diff
@@ -0,0 +1,43 @@
...
+    skip: Property[bool] = Property.default(False)
```
ntrinquier: Why do we need a skip property? If the caller is able to determine they do not want the task to run when creating it, they might as well not create it, no?

kraken-std/src/kraken/std/cargo/tasks/cargo_generate_deb.py
morosanmihail: That is a good shout!

ntrinquier (approved):

PR #89, opened 2023-07-27
Add log level property

ddavid:
Enables to set the `--log-cli-level` via a task property.

### DIFF
--- a/kraken-std/src/kraken/std/python/tasks/pytest_task.py
+++ b/kraken-std/src/kraken/std/python/tasks/pytest_task.py
@@ -12,6 +12,13 @@ from kraken.core import Project, Property, TaskStatus
 from .base_task import EnvironmentAwareDispatchTask
 
 
+class LogLevel(enum.Enum):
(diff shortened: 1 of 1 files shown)


kraken-std/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -36,6 +43,7 @@ class PytestTask(EnvironmentAwareDispatchTask):
     doctest_modules: Property[bool] = Property.default(True)
     marker: Property[str]
     coverage: Property[CoverageFormat]
+    log_level: Property[LogLevel] = Property.default(LogLevel.INFO)
```
ntrinquier, line 46: You can use [the levels from the standard library](https://docs.python.org/3/library/logging.html#logging-levels), and then type this as:

```suggestion
    log_level: Property[Literal[logging.DEBUG, logging.INFO, ...]] = Property.default(logging.INFO)
```

ntrinquier:
@ddavid do you still want this in?

PR #90, opened 2023-07-27
Allow specifying additional directories for Python linting

cowlingjosh:
Often projects contain an `examples/` or `bin/` directory which contains Python source code, but by default Kraken only lints against source code in `src` and `tests` directories.

---

This MR adds the ability to specify any number of extra directories for which lint tasks (`black`, `flake8`, `isort`, `mypy`, `pycln`, `pylint`) should also be run against. This functionality is exposed via the `python_settings` function which modifies the global `PythonSettings` object.


kraken-std/pyproject.toml
```diff
@@ -40,6 +40,7 @@ isort = "*"
 mypy = "*"
 pycln = "*"
 pytest = "*"
+pylint = "*"
```
ntrinquier:
```suggestion
pylint = "*"
pytest = "*"
```

kraken-std/src/kraken/std/python/settings.py
```diff
@@ -139,6 +140,7 @@ def python_settings(
     build_system: PythonBuildSystem | None = None,
     source_directory: str | Path | None = None,
     tests_directory: str | Path | None = None,
+    additional_directories: list[str | Path] | None = None,
```
ntrinquier: I'm not a big fan of the name ("additional" is vague and non-descriptive), but I don't have a better suggestion... I'll think about it.

kraken-std/src/kraken/std/python/settings.py
```diff
@@ -178,6 +182,13 @@ def python_settings(
     if tests_directory is not None:
         settings.tests_directory = Path(tests_directory)
 
+    if additional_directories is not None:
+        settings.additional_directories = [
+            path
+            for additional_directory in additional_directories
+            if (path := Path(additional_directory)).exists() and path.is_dir()
```
ntrinquier: It won't give you a neat one-liner, but perhaps we should print paths which do not exist or are not directories?

ntrinquier:
@cowlingjosh is this still relevant?

kraken-std/src/kraken/std/python/settings.py
cowlingjosh: Naming is hard! Maybe something more descriptive about how these are directories are just being linted? Something like `lint_enforced_directories`?

kraken-std/src/kraken/std/python/settings.py
cowlingjosh: Good shout, done :)

cowlingjosh:
> @cowlingjosh is this still relevant?

Apologies for the late reply on this @ntrinquier, I missed the email notifications from Github 🤦 Thanks for the review, I have added your suggestions.

kraken-std/src/kraken/std/python/settings.py
```diff
@@ -39,6 +39,7 @@ class _PackageIndex(PackageIndex):
     build_system: PythonBuildSystem | None = None
     source_directory: Path = Path("src")
     tests_directory: Path | None = None
+    lint_enforced_directories: list[Path] | None = None
```
ntrinquier: As it is a sequence, we can make it non nullable and default to an empty sequence: it will make code in the tasks a bit more simple because then we can just always iterate through `lint_enforced_directories` without having to check for `None` first.

```suggestion
    lint_enforced_directories: list[Path] = dataclasses.field(default_factory=list)
```

By doing that we lose the ability to distinguish between "this value has been provided but it is empty" and "this value has not been provided", but I think this is acceptable (and arguably: desirable :)).

kraken-std/src/kraken/std/python/tasks/black_task.py
```diff
@@ -25,6 +25,8 @@ class BlackTask(EnvironmentAwareDispatchTask):
     def get_execute_command(self) -> list[str]:
         command = ["black", str(self.settings.source_directory)]
         command += self.settings.get_tests_directory_as_args()
+        if self.settings.lint_enforced_directories is not None:
+            command += [str(directory) for directory in self.settings.lint_enforced_directories]
```
ntrinquier: If `lint_enforced_directories` is not nullable, then this (and other tasks) can be simplified as

```suggestion
        command += [str(directory) for directory in self.settings.lint_enforced_directories]
```

ntrinquier (commented):
@cowlingjosh one last small comment, let me know what you think!

kraken-std/src/kraken/std/python/settings.py
cowlingjosh: I like this, no reason we can't be more specific with the `PythonSettings` object compared to with `python_settings`. Thanks for the suggestion, I have made the change.

cowlingjosh:
@ntrinquier thanks again for the suggestions, I have made the changes. 

I have made two other changes as well:
- Updated the `poetry.lock` file of the `kraken-wrapper` subproject after the recent `^0.31.0` version bump.
- Changed the log level of the _directory does not exist_ statements to `DEBUG`, as most likely we want some defaults (such as examples/ etc.) and do not want to overload the output for when these directories do not exist (which will be often).

PR #91, opened 2023-07-31
Fixes Flake8 lint by using `is not` instead of `!=`

Tpt:
We are already checking type of other...

### DIFF
--- a/kraken-common/src/kraken/common/supplier.py
+++ b/kraken-common/src/kraken/common/supplier.py
@@ -136,7 +136,7 @@ class MapSupplier(Supplier[U], Generic[T, U]):
         return f"{self._value}.map({self._func})"
 
     def __eq__(self, other: object) -> bool:
-        if type(other) != type(self):
+        if type(other) is not type(self):
             return False
         assert isinstance(other, MapSupplier)
         return (self._func, self._value) == (other._func, other._value)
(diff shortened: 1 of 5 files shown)


kraken-common/src/kraken/common/supplier.py
```diff
@@ -136,8 +136,6 @@ def __repr__(self) -> str:
         return f"{self._value}.map({self._func})"
 
     def __eq__(self, other: object) -> bool:
-        if type(other) != type(self):
-            return False
```
ntrinquier: That's not equivalent to the current behaviour because 1. this raises an `AssertionError` instead of returning `False`, and 2. `isinstance` is looser than strict type checking (because it matches inherited objects as well).

I think the proper way to fix the flake8 error is to replace the type check with 

```python
        if type(self) is not type(other):
            return False
```

(using `is not` instead of `!=`)

Tpt: Indeed. my bad. Done.

ntrinquier (approved):

ntrinquier:
@Tpt There are more apparently https://github.com/kraken-build/kraken-build/blob/d67179a8f2a93d409cfe3288c967119bfa503cb1/kraken-core/src/kraken/core/system/context_test.py#L47-L54

(Sadly they are not reported at the same time when flake8 fails)

Tpt:
> @Tpt There are more apparently
> 
> https://github.com/kraken-build/kraken-build/blob/d67179a8f2a93d409cfe3288c967119bfa503cb1/kraken-core/src/kraken/core/system/context_test.py#L47-L54
> 
> (Sadly they are not reported at the same time when flake8 fails)

Fixed!

PR #92, opened 2023-08-02
Report python coverage in terminal

bpoumarede:


kraken-std/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -61,6 +61,8 @@ def get_execute_command(self) -> list[str] | TaskStatus:
             command += [
                 "--cov-report",
                 f"{self.coverage.get().get_format()}:{str(self.project.build_directory / coverage_file)}",
+                "--cov-report",
+                "term",
```
ntrinquier, line 65: Why is this needed / can we make this configurable?

bpoumarede: It is available only if coverage is enabled. It is needed for gitlab to print the coverage in the MR.

ntrinquier (approved):

PR #94, opened 2023-08-02
Correct multi-line description wrapping

### LINKED ISSUE #93: Incorrect line wrapping for tasks with long descriptions
To reproduce:

1. Define a task whose description is long enough to wrap in your terminal.
2. Make your terminal fairly narrow.
3. Run `kraken q ls`.
4. Notice that the wrapped lines of the description (that is, every line except the first for each task) is wrapped _very_ narrowly.

This happens because we use the `subsequent_ident` argument to `textwrap.wrap`:

https://github.com/kraken-build/kraken-build/blob/fcb55055a56ec703729315b462820cfd363c26bb/kraken-core/src/kraken/core/cli/main.py#L324-L328

And `subsequent_indent` [counts towards the line length of each line except the first](https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.subsequent_indent). In other words, if the terminal is 120 wide and `remaining_width = 60`, we'll end up passing 120-60 space characters to `subsequent_lines`, and those will be counted against the `remaining_width` we pass as the `width` argument. Which effectively means there is _zero_ remaining space, and so `textwrap` dutifully wraps the remaining lines as aggressively as it can.

The solution here is probably to _not_ pass `subsequent_indent`, and instead just prepend the spaces onto each line (except the first) that `textwrap.wrap` returns.

jonhoo:
Fixes #93.

### DIFF
--- a/kraken-core/src/kraken/core/cli/main.py
+++ b/kraken-core/src/kraken/core/cli/main.py
@@ -321,11 +321,11 @@ def ls(graph: TaskGraph) -> None:
             remaining_width -= 2
             if remaining_width <= 0:
                 remaining_width = width
-            for part in textwrap.wrap(
-                description,
-                remaining_width,
-                subsequent_indent=(width - remaining_width) * " ",
-            ):
+            parts = textwrap.wrap(description, remaining_width)
+            line.append(parts[0]) # the first line is already indented
+            line.append("\n")
+            for part in parts[1:]:
+                line.append((width - remaining_width - 1) * " ")
                 line.append(part)
                 line.append("\n")
             line.pop()
--- a/kraken-core/src/kraken/core/cli/main.py
+++ b/kraken-core/src/kraken/core/cli/main.py
@@ -322,7 +322,7 @@ def ls(graph: TaskGraph) -> None:
             if remaining_width <= 0:
                 remaining_width = width
             parts = textwrap.wrap(description, remaining_width)
-            line.append(parts[0]) # the first line is already indented
+            line.append(parts[0])  # the first line is already indented
             line.append("\n")
             for part in parts[1:]:
                 line.append((width - remaining_width - 1) * " ")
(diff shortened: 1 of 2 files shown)


jonhoo:
Hmm, I tried adding a changelog entry with `slap changelog add`, but got the error:
```
error: cannot add changelog because the feature must be enabled in the config
```
🤔

kraken-core/src/kraken/core/cli/main.py
```diff
@@ -321,13 +321,16 @@ def _print_task(task: Task) -> None:
             remaining_width -= 2
             if remaining_width <= 0:
                 remaining_width = width
+            first = True
             for part in textwrap.wrap(
```
tetigi: Nit: not a huge fan of boolean-flags within loop state. Perhaps we can be a bit more obvious with something like:

```python
parts = textwrap.wrap(description, remaining_width)
line.append(part[0]) # we only want to indent lines after the first
line.append("\n")
for part in parts[1:]:
    line.append((width - remaining_width - 1) * " ")
    line.append(part[0])
    line.append("\n")   
```

Just a style-choice overall, so not a blocker.

tetigi (approved):
One nit; leave styling choice to your discretion ;)

tetigi:
> Hmm, I tried adding a changelog entry with `slap changelog add`, but got the error:
> 
> ```
> error: cannot add changelog because the feature must be enabled in the config
> ```
> 
> 🤔

Likely you ran it from the root dir, and not the sub-dir (or vice versa)?

kraken-core/src/kraken/core/cli/main.py
jonhoo: Good call! Changed it to match this :)

jonhoo:
That was exactly it, thanks! Force-pushed so that we get a single nicely-formatted commit.

kraken-core/src/kraken/core/cli/main.py
NiklasRosenstein: Another good way to solve this can be `for idx, part in enumerate(parts):` and checking for `idx == 0`. Though better suited if the code you would otherwise need to replicate in @tetigi's version is a but more. Here it definitely improves clarity to move the zero-index case out of the loop!

PR #95, opened 2023-08-03
Allow to pass a custom error message for cargo_deny

ntrinquier:


kraken-std/src/kraken/std/cargo/tasks/cargo_deny_task.py
```diff
@@ -19,4 +20,11 @@ def execute(self) -> TaskStatus:
...
 
         result = subprocess.run(command, cwd=self.project.directory)
-        return TaskStatus.from_exit_code(command, result.returncode)
+        if result.returncode == 0:
+            return TaskStatus.succeeded()
+
+        return (
+            TaskStatus.failed(self.error_message.get())
+            if self.error_message.get() is not None
+            else TaskStatus.from_exit_code(command, result.returncode)
```
NiklasRosenstein, line 29: A `.map()` might be better suited here and would help to avoid calling `get()` twice.

ntrinquier: https://github.com/kraken-build/kraken-build/pull/97

PR #98, opened 2023-08-04
kraken-core/: fix: Rework task description wrapping

jonhoo:
This is a follow-up to #94. It started because I noticed a boundary-condition bug where descriptions that came all the way up to right side of the terminal seemed to have an extra newline added to them. That turned out to be because every line got an extra space added to it, so all the lines ended up one character too long, and then regular terminal wrapping kicked in.

This caused me to dig further into how exactly the description-rendering code was set up, which led me to find some more oddities. So I decided to make two bigger changes to the code.

First, it no longer joins `line` by spaces, but instead joins by `\n`, with every string-append to a line prefixing its own separator (`+= " "`). This makes it easier to determine where line breaks actually come from, and avoids spurious spaces from appearing.

And second, if there is so little room remaining on a line that it would look weird when wrapped (see code comment for how that's evaluated), the description is now moved to the next line and wrapped independently to 80 characters wide. If this happens, extra vertical spacing is also added before the next task.

I've tested this by adjusting the description and task status of a number of tasks and running `kraken q ls` to verify that the output (subjectively) looks readable and sane. That's also how I landed on 48 characters as the "right" place to flip between column and next-line mode.

Some screenshots:
Screenshot 2023-08-04 at 11 00 33
Screenshot 2023-08-04 at 11 01 55
Screenshot 2023-08-04 at 11 02 25

### DIFF
--- a/kraken-core/.changelog/_unreleased.toml
+++ b/kraken-core/.changelog/_unreleased.toml
@@ -3,6 +3,7 @@ id = "cb501ed7-e633-4a73-bd0e-0e95bdeeb96d"
 type = "fix"
 description = "Rework task description wrapping"
 author = "jon.gjengset@helsing.ai"
+pr = "https://github.com/kraken-build/kraken-build/pull/98"
 issues = [
     "https://github.com/kraken-build/kraken-build/issues/93",
 ]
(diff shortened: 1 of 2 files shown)


ntrinquier (approved):
Thanks @jonhoo!

PR #108, opened 2023-09-13
Fixes CI: enforces isort and updates poetry.lock

Tpt:


alexespencer (approved):
Thanks - looks good and I had a similar change for the isort on that file on an unreleated PR, but this is a better place for it, along with your fixes

NiklasRosenstein (approved):

PR #109, opened 2023-09-13
Asp/move http to common

alexespencer:
For an upcoming change to `kraken-wrapper` that requires the use of http, the module sits better in `kraken-common` than `kraken-std`. For now, I've copied the module with the linting tests, and left the module in `kraken-std` but added a warning that it's deprecated


alexespencer:
Note: I did not write the http module, I've lifted it from `kraken.std` - there were previous comments on that PR about the function names, particularly within the linting section, but I've chosen to keep them as is, lift and shift.

kraken-common/pyproject.toml
```diff
@@ -60,7 +62,7 @@ flake8 = "flake8 src/"
 fmt = "black src/ && isort src/"
 
 [tool.mypy]
-python_version = "3.6"
+python_version = "3.10"
```
NiklasRosenstein: You'll want to keep this at the minimum supported Python version, which is 3.7 (see above). :)

kraken-common/src/kraken/common/http/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+
+def patch(url: str, **kwargs: Any) -> httpx.Response:
+    return httpx.patch(url, verify=_get_system_ca_list(), **kwargs)
+
+
+def delete(url: str, **kwargs: Any) -> httpx.Response:
+    return httpx.delete(url, verify=_get_system_ca_list(), **kwargs)
+
+
+# Forwarding calls to `stream` would not be accepted as-is by mypy.
```
NiklasRosenstein: Did this not work?

```py
def stream(url: str, **kwargs: Any) -> ContextManager[httpx.Response]:
    return httpx.stream(url, verify=_get_system_ca_list(), **kwargs)
```

NiklasRosenstein (commented):
Left two comments, but overall LGTM 🚀

kraken-common/pyproject.toml
alexespencer: Done, thanks

kraken-common/src/kraken/common/http/__init__.py
alexespencer: In short, no. Error below...shall I leave as is?

<details>
  <summary>Error</summary>
kraken-common:mypy| src/kraken/common/http/__init__.py:67: error: Argument 1 to "contextmanager" has incompatible type
kraken-common:mypy| "Callable[[str, KwArg(Any)], AbstractContextManager[Response]]"; expected
kraken-common:mypy| "Callable[[str, KwArg(Any)], Iterator[<nothing>]]"  [arg-type]
</details>

NiklasRosenstein: Is it a typo or something that `contextmanager` is lowercase? Are you using `typing.ContextManager`?

alexespencer: I believe it's picking that up from the annotation. I definitely used typing.ContextManager:

```
from typing import Any, ContextManager
...
@contextmanager
def stream(url: str, **kwargs: Any) -> ContextManager[httpx.Response]:
    return httpx.stream(url, verify=_get_system_ca_list(), **kwargs)`
```

The mypy error (From CI) is the same...

NiklasRosenstein: Aha! When you're already returning a context manager, you don't need to wrap the function with `@contextmanager`. :) The decorator is an adapter for a function with the signature `(...) -> Iterator[T]` to `(...) -> ContextManager[T]`. But your function already returns a `ContextManager`. So if you just get rid of the decorator, I believe it should work.

NiklasRosenstein (approved):

PR #112, opened 2023-09-14
Add masking. Add auth check

alexespencer:
This change:

- [x] masks password output when running `krakenw auth --list` **by default**
- [x] adds a `--no-mask` switch to display the plain text password
- [x] performs a credential check for jfrog.io and github by default, unless `--no-check` is provided. Most users are tripping up here and having credential check automatically performed (vs requesting it) will help
- [x] displays `curl` commands for users to run themselves if they want to, requested via `--verbose`
- [x] displays common warnings based upon the response received from JFrog (like using an API token vs an identity token)

Users can edit ~/krakenw/config.toml if they would like a different URL to be used for either jfrog or github:
```
[auth."helsing.jfrog.io"]
username = "alex.spencer@helsing.ai"
auth_check_url_suffix = "artifactory/api/pypi/python-all/simple/flask/"
```

For future changes, we'll need to extend the auth abilities, and may want to consider refactoring this so it's easier to maintain


kraken-wrapper/src/kraken/wrapper/main.py
```diff
@@ -176,6 +180,33 @@ def auth(prog: str, argv: list[str], use_keyring_if_available: bool) -> NoReturn
...
+
+                # If verbose, also display the CURL command that people can use plus the first part of the response
+                if args.verbose:
+                    logger.info("Checking auth for host %s with command: %s", host, credential_result.curl_command)
+                    logger.info(
+                        "First 10 lines of response (limited to 1000 chars): %s",
+                        ("\n".join(credential_result.raw_result.split("\n")[0:10])[0:1000]),
+                    )
+        except ReadTimeout:
+            logger.warning("HTTP Timeout when testing credentials")
```
Tpt: This exception is raised by the http call in `check_credential`, it might be clearer to catch it there

alexespencer: Moved, cleans up main.py a lot (1 less import) and using the exception closer to where I expect it is clearer - appreciate the suggestion, thank you

NiklasRosenstein (approved):

PR #120, opened 2023-09-28
Add buffrs login, install, publish and generate to kraken-std

alexespencer:


kraken-std/src/kraken/std/buffrs/manifest.py
```diff
@@ -0,0 +1,74 @@
...
+        cloned = dict(json)
+        type_ = cloned.pop("type")
+        name = cloned.pop("name")
+        version = cloned.pop("version")
+        return Package(type_, name, version, cloned)
+
+    def to_json(self) -> dict[str, str]:
+        values = {"type": self.type_, "name": self.name, "version": self.version}
+
+        if self.unhandled is not None:
```
NiklasRosenstein: Can not be None as per its type signature

kraken-std/src/kraken/std/buffrs/manifest.py
```diff
@@ -0,0 +1,74 @@
...
+            values.update({k: v for k, v in self.unhandled.items() if v is not None})
```
NiklasRosenstein: Why filter the values by `None`?

kraken-std/src/kraken/std/buffrs/tasks.py
```diff
@@ -0,0 +1,138 @@
...
+
+class BuffrsLoginTask(Task):
+    """This task logs into artifactory with buffrs"""
+
+    description = "Login to artifactory with buffrs."
+    artifactory_credentials: Property[CredentialsWithHost]
+
+    def execute(self) -> TaskStatus:
+        credentials = self.artifactory_credentials.get()
+        url = urljoin(credentials.host, "/artifactory")
```
NiklasRosenstein: I would suggest that we keep what happens between the task and the buffrs command it runs as transparent as possible. Instead of appending `/artifactory`, we should expect the URL to already contain it.

kraken-std/src/kraken/std/buffrs/tasks.py
```diff
@@ -0,0 +1,138 @@
...
+
+        if manifest.package is None:
+            return manifest.to_toml_string()
+
+        manifest.package.version = self.version.get().format()
+
+        return manifest.to_toml_string()
+
+    def start_background_task(self, exit_stack: contextlib.ExitStack) -> TaskStatus | None:
+        project = self.project or Project.current()
```
NiklasRosenstein: `self.project` is never `None`

kraken-std/.changelog/_unreleased.toml
```diff
@@ -0,0 +1,6 @@
+[[entries]]
+id = "aa631280-287e-4773-9118-89619e9bd5e7"
+type = "feature"
+description = "Move buffrs login/install/generate to kraken.std"
```
NiklasRosenstein: nit: The public doesn't understand where we moved it
from so maybe we just say "added ..."

NiklasRosenstein (changes requested):

kraken-std/src/kraken/std/buffrs/manifest.py
alexespencer: I'm not familiar with the manifest.py section so I'll look into this 👍

alexespencer: I've removed the filter - you can serialise None - I can't see historic reasoning / comments - it may have been to reduce the size of the JSON doc?

kraken-std/src/kraken/std/buffrs/tasks.py
```diff
@@ -0,0 +1,136 @@
...
+            command,
+            sp.call(
+                command,
+                cwd=self.project.directory,
+                env=os.environ.copy(),
+            ),
+        )
+
+
+class BuffrsBumpVersionTask(BackgroundTask):
```
NiklasRosenstein: @alexespencer I vaguely remember a thread on buffrs about not having the version number in the Buffrs manifest, I just want to confirm that buffrs still does store the version there?

If `buffrs publish` would accept a `--version` argument, this temporary version bump would not be necessary.

NiklasRosenstein: If we do still need the version bump in the file, I would suggest we just make that logic part of the publish task. I don't think there's any other task that requires the version to be bumped?

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -357,6 +363,7 @@ def cargo_test(
     group: str | None = "test",
     project: Project | None = None,
     features: list[str] | None = None,
+    depends_on: Sequence[Task] | None = None,
```
NiklasRosenstein:
```suggestion
    depends_on: Sequence[Task] = (),
```

Then you don't need the `if` below.

kraken-std/src/kraken/std/cargo/__init__.py
```diff
@@ -306,6 +306,7 @@ def cargo_build(
     name: str | None = None,
     project: Project | None = None,
     features: list[str] | None = None,
+    depends_on: Sequence[Task] | None = None,
```
NiklasRosenstein:
```suggestion
    depends_on: Sequence[Task] = (),
```

Then you don't need the `if` below.

NiklasRosenstein: Can you provide a rationale in the changelog for this change?

kraken-std/src/kraken/std/buffrs/tasks.py
alexespencer: Hey Niklas, `buffrs publish` does not support `--version` - so I'll move this login into the publish task

NiklasRosenstein: Just to clarify, the login is probably good to have as a separate task. People can run `krakenw run buffrs.login` to explicitly propagate their Buffrs credentials. :) This is about the version bump.

kraken-std/src/kraken/std/buffrs/tasks.py
```diff
@@ -0,0 +1,130 @@
...
+            command,
+            sp.call(
+                command,
+                cwd=self.project.directory,
+                env=os.environ.copy(),
+            ),
+        )
+
+
+class BuffrsPublishTask(BackgroundTask):
```
NiklasRosenstein: This doesn't need to be a background task now. You just need to move the `atomic_file_swap()` call into the `execute()` method and wrap it around the Buffrs command invokation.

NiklasRosenstein (approved):

NiklasRosenstein:
🚀

PR #130, opened 2023-11-08
kraken-build/: improvement: Calling `buildscript()` from a Kraken project that is not the root project will now raise a `RuntimeError`.

### LINKED ISSUE #124: Raise an error if `buildscript()` is called not from the root project
People that are not very familiar with Kraken that write a `.kraken.py` for a sub-project will often copy the `buildscript()` block into that script, but it's only required at the top-level. We should just raise an exception when it is used in any project that is not the root project, as it doesn't add any value there.

NiklasRosenstein:


NiklasRosenstein:
Closes #124

PR #135, opened 2023-12-05
Add workspace flag to cargo_test

asmello:
This flag helps ensuring that all tests across all crates are applied, and not just the default ones. It matches the same flag that already exists in `cargo_build`.

### DIFF
--- a/kraken-build/src/kraken/std/cargo/__init__.py
+++ b/kraken-build/src/kraken/std/cargo/__init__.py
@@ -363,6 +363,7 @@ def cargo_test(
     project: Project | None = None,
     features: list[str] | None = None,
     depends_on: Sequence[Task] = (),
+    workspace: bool | None = None,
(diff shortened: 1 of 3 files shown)


NiklasRosenstein (approved):

kraken-build/src/kraken/std/cargo/__init__.py
```diff
@@ -363,14 +363,16 @@ def cargo_test(
     project: Project | None = None,
     features: list[str] | None = None,
     depends_on: Sequence[Task] = (),
+    workspace: bool | None = None,
```
wngr, line 366: Shouldn't it default to `True`?

asmello: That'd be a semantic break here, and inconsistent with `cargo_build()`. We can set our opinionated default internally.

PR #136, opened 2023-12-11
Use non-strict mode when parsing .gitconfig

qsantos:
I recognize that my `~/.gitconfig` is a bit messy, but it has never been a problem before. This PR ensure that it works with kraken-build as well, removing a potential friction for users.

By default, configparser will raise when encountering the same section several times. There is no reason to do so here, and we can just merge the occurrences. This is done by passing strict=False to the configuration parser.


NiklasRosenstein (approved):
Sounds reasonable. Thanks! 👍

qsantos:
I have a failure with isort in the CI (ran thrice to be sure), but not locally. Is it possible that the isort version differs between the context, or is its version pinned?

NiklasRosenstein:
Just had to merge in develop :)

PR #138, opened 2023-12-11
Fix `kraken run -X <task>` option to correctly not mark a required task as skipped

NiklasRosenstein:
- fmt
- kraken-build/: tests: Add a unit test for a real-world example of a DAG that failed the expected behavior of `TaskGraph.mark_tasks_as_skipped()`.
- kraken-build/: fix: Replace `Tasks.mark_tasks_as_skipped()` with a simpler implementation that also fixes an issues with second-degree tasks being marked as skipped even if they should not be.


NiklasRosenstein:
Something we should consider eventually as a follow up: If you mark a task as skipped that is required to run for a successor task (e.g. if it provides an output property), the subsequent task _will_ fail. The error message will be a bit obscure in this case (e.g. "Property value will be known at a later time")

PR #140, opened 2023-12-13
Enable cargo:token as a global-credential-provider

qsantos:
Enable cargo:token as a global-credential-provider

Since 1.74, Cargo requires a credential provider to be set up to download crates from private registries. This changes set the cargo:token credential provider. This stays compatible with the mitmproxy credentials injection.

This will also be useful when switching to cargo-based authentication instead of relying on credentials injection.

### DIFF
--- a/kraken-build/.changelog/_unreleased.toml
+++ b/kraken-build/.changelog/_unreleased.toml
@@ -1,0 +1,6 @@
+[[entries]]
+id = "a0755acd-c588-47a7-bd6d-7ade753d1d1e"
+type = "fix"
+description = "Enable cargo:token as a global-credential-provider; this is required since 1.74, which brings authenticated priviate registries"
+author = "quentin.santos@helsing.ai"
+pr = "https://github.com/kraken-build/kraken/pull/140"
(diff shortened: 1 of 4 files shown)


kraken-build/.changelog/_unreleased.toml
```diff
@@ -0,0 +1,6 @@
+[[entries]]
+id = "a0755acd-c588-47a7-bd6d-7ade753d1d1e"
+type = "fix"
+description = "Enable cargo:token as a global-credential-provider"
```
NiklasRosenstein: Could we add the rationale here please? Something along the lines of "this is required for Cargo 0.74.0 and up because .."

NiklasRosenstein (changes requested):

kraken-build/.changelog/_unreleased.toml
qsantos: Done!

NiklasRosenstein (commented):
🚀

NiklasRosenstein (approved):

PR #144, opened 2023-12-18
Adds cargo hack as optional task

xfbs:
Since <https://github.com/kraken-build/kraken-base-image/pull/60> added `cargo-hack` to the tools available in the base image, the next step is to expose it cleanly for use by kraken jobs.

This PR aims to do just that, exposing a `cargo_hack()` API which can be used to add a `cargoHack` task to verify dependencies.

It can be enabled like this in a `.kraken.py` file:

```
cargo_hack()
```

Optionally, it can also be given arguments:

```
cargo_hack(features = CargoHackFeatures.POWERSET, action = CargoHackAction.BUILD)
```

This PR also comes with some unit tests to test the thing because I don't trust my python skills.


kraken-build/pyproject.toml
```diff
@@ -18,7 +18,7 @@ packages = [
   {include = "kraken/std", from = "src"},
 ]
 readme = "README.md"
-version = "0.32.4"
+version = "0.32.4.post0.dev2+dirty"
```
NiklasRosenstein: Undo those please :)

You don't need to run the `build` task for local dev, `fmt lint test` is enough

xfbs: Sure! Honestly I don't know what I'm doing, I saw those in the diff before making the commit and I already figured that they don't belong 😛

xfbs: Done — just pushed a commit. I did a `git rebase`, `git reset --soft HEAD~1` and `git restore`'d every file which contains a dirty version (some `pyproject.toml` files, some `__init__.py` files). Hope that is right!

kraken-build/src/kraken/std/cargo/__init__.py
```diff
@@ -232,6 +237,42 @@ def cargo_deny(
...
+) -> CargoHackTask:
+    """Adds a task running cargo-hack for cargo projects. This ensures that all
+    combinations of the features declared in the Cargo manifests build
+    cleanly.
+
+    :param features: Method to select the features to test for, possible values
+    are CargoHackFeatures.EACH and CargoHackFeatures.POWERSET.
+    :param action: Action to run for every selected set of features, possible
+    values are CargoHackAction.BUILD, CargoHackAction.CHECK and
+    CargoHackAction.TEST.
```
NiklasRosenstein:
```suggestion
    :param features: Method to select the features to test for, possible values
        are CargoHackFeatures.EACH and CargoHackFeatures.POWERSET.
    :param action: Action to run for every selected set of features, possible
        values are CargoHackAction.BUILD, CargoHackAction.CHECK and
        CargoHackAction.TEST.
```

kraken-build/src/kraken/std/cargo/__init__.py
```diff
@@ -232,6 +237,42 @@ def cargo_deny(
     )
 
 
+def cargo_hack(
+    *,
+    project: Project | None = None,
+    features: CargoHackFeatures = CargoHackFeatures.EACH,
```
NiklasRosenstein: For usability within `.kraken.py` it would be useful if this also accepted a string! So I can use `cargo_hack(features="powerset")` which is easier to type, read and remember. In case you didn't know, `Enum` types have a `__getitem__` support so that's straight forward to implement :)

kraken-build/src/kraken/std/cargo/__init__.py
```diff
@@ -232,6 +237,42 @@ def cargo_deny(
...
+
+    project = project or Project.current()
+    task = project.do(
+        "cargoHack" if name is None else name,
+        CargoHackTask,
+        features=features,
+        action=action,
+        error_message=error_message,
+        group=group,
+    )
```
NiklasRosenstein:
```suggestion
    task = project.task(name, CargoHackTask)
    task.features = features
    task.action = action
    task.error_message = error_message
    task.group = group
```

`Project.do()` is actually deprecated in favor of `Project.task()`. This is type safe as opposed to the `do()` variant!

If you give `name` a default value you don't need to add the `is None` check.

kraken-build/src/kraken/std/cargo/tasks/cargo_hack_task.py
```diff
@@ -0,0 +1,57 @@
...
+    action_arguments: Property[list[str]] = Property.default_factory(list)
+
+    def execute(self) -> TaskStatus:
+        command = ["cargo", "hack"]
+
+        match self.features.get():
+            case CargoHackFeatures.EACH:
+                command.append("--each-feature")
+            case CargoHackFeatures.POWERSET:
+                command.append("--feature-powerset")
```
NiklasRosenstein, line 35:
```suggestion
                command.append("--feature-powerset")
            case _: assert False, f"invalid features={self.features.get()}"
```

kraken-build/src/kraken/std/cargo/tasks/cargo_hack_task.py
```diff
@@ -0,0 +1,57 @@
...
+
+        match self.action.get():
+            case CargoHackAction.CHECK:
+                command.append("check")
+            case CargoHackAction.TEST:
+                command.append("test")
+            case CargoHackAction.BUILD:
+                command.append("build")
```
NiklasRosenstein, line 45:
```suggestion
                command.append("build")
            case _: assert False, f"invalid action={self.action.get()}"
```

kraken-build/src/kraken/std/cargo/tasks/cargo_hack_task.py
```diff
@@ -0,0 +1,57 @@
...
+
+        for arg in self.action_arguments.get():
+            command.append(arg)
```
NiklasRosenstein:
```suggestion
        command += self.action_arguments.get()
```

kraken-build/src/kraken/std/cargo/tasks/cargo_hack_task.py
```diff
@@ -0,0 +1,57 @@
...
+
+        result = subprocess.run(command, cwd=self.project.directory)
+        if result.returncode == 0:
+            return TaskStatus.succeeded()
+
+        return self.error_message.map(
```
NiklasRosenstein: Where do you expect `error_message` to be filled? From the user in `.kraken.py`? What would a possible error message be?

I lead on removing the property and just returning `TaskStatus.from_exit_code(command, result.returncode)`.

NiklasRosenstein (commented):
A few suggestions on the code, but overall this looks good to me! Nice job on the unit tests as well 🙌

kraken-build/src/kraken/std/cargo/__init__.py
xfbs: Done — I've made it accept either types and used `isinstance()` to figure out the type at runtime, to convert it. Hope that works :)

kraken-build/src/kraken/std/cargo/tasks/cargo_hack_task.py
xfbs: Honestly, this code is copy paste from another task and I did not know what I'm trying to do :D will do what you suggest.

xfbs:
Thank you for the feedback! I've addressed all of the suggestions :)

NiklasRosenstein:
Closing because this didn't really come up as a requested feature again ever since; happy to reconsider though.

PR #148, opened 2023-12-19
Vendor `dill` for Python 3.11+ support

NiklasRosenstein:


NiklasRosenstein:
No longer needed as Dill finally released the new version.

PR #153, opened 2023-12-21
New command cargoLogin to login in all cargo registries

qsantos:
It seems to work locally. Hopefully, it will also work in the CI.


kraken-build/src/kraken/std/cargo/tasks/cargo_sync_config_task.py
```diff
@@ -36,10 +38,33 @@ def __init__(self, name: str, project: Project) -> None:
 
     def get_file_contents(self, file: Path) -> str | bytes:
         content = tomli.loads(file.read_text()) if not self.replace.get() and file.exists() else {}
-        content.setdefault("registry", {})["global-credential-providers"] = ["cargo:token"]
+        content.setdefault("registry", {})["global-credential-providers"] = [
+            "cargo:libsecret",
+            "cargo:macos-keychain",
+            "cargo:wincred",
+            "cargo:token",
```
ntrinquier: Can you leave a comment explaining the rationale for the order?

kraken-build/src/kraken/std/cargo/tasks/cargo_sync_config_task.py
```diff
@@ -36,10 +38,33 @@ def __init__(self, name: str, project: Project) -> None:
...
+            )
+            if p.returncode != 0:
+                if p.stderr.endswith(b"\nerror: config.json not found in registry\n"):
+                    # expected error, ignore
+                    pass
+                else:
+                    # unknown error, fail normally
+                    stderr.write(p.stderr.decode())
+                    p.check_returncode()
+
```
ntrinquier, line 51: I think it would be more appropriate to have the login happen in a separate task, `cargoLogin`:

- `cargoLogin` would depend on `cargoSyncConfig`
- `cargoLogin` would exist in two variants, `cargoLoginBuild` and `cargoLoginPublish` respectively belonging to the groups `cargoBuildSupport` and `cargoPublishSupport`, and using the read and publish tokens

This way we separate concerns (one tasks syncs the config file, another task logins), and we can ensure the proper tokens are used for each operation.

kraken-build/src/kraken/std/cargo/tasks/cargo_sync_config_task.py
```diff
@@ -36,10 +38,33 @@ def __init__(self, name: str, project: Project) -> None:
...
             content.setdefault("registries", {})[registry.alias] = {"index": registry.index}
+            p = run(
+                ["cargo", "login", "--registry", registry.alias],
+                cwd=self.project.directory,
+                capture_output=True,
+                input=publish_token.encode(),
+            )
+            if p.returncode != 0:
+                if p.stderr.endswith(b"\nerror: config.json not found in registry\n"):
+                    # expected error, ignore
```
ntrinquier: Why is it expected?

kraken-build/src/kraken/std/cargo/tasks/cargo_sync_config_task.py
NiklasRosenstein: Agree on the above, just one question: Do we need to differentiate between build and publish tokens?

ntrinquier: Apart from the obvious fact it enables granular permissions management (whose benefits are debatable in this context), I think the current distinction also exists because the publish token is passed to the [`cargo publish --token ...` command](https://github.com/kraken-build/kraken/blob/5b1babf9f9419b0cb7da25036176c62699207796/kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py#L39), whereas the read credentials are [injected by the MITM proxy using basic auth](https://github.com/kraken-build/kraken/blob/5b1babf9f9419b0cb7da25036176c62699207796/kraken-build/src/kraken/std/cargo/tasks/cargo_auth_proxy_task.py#L100) (so it also requires a username, but the token could very well be the same).

If we decide we do not need granular permissions, we could replace [`read_credentials`](https://github.com/kraken-build/kraken/blob/5b1babf9f9419b0cb7da25036176c62699207796/kraken-build/src/kraken/std/cargo/config.py#L22) and [`publish_token`](https://github.com/kraken-build/kraken/blob/5b1babf9f9419b0cb7da25036176c62699207796/kraken-build/src/kraken/std/cargo/config.py#L25) with a single `token` property, and only rely on `cargo login` to make the token available for read and publish.

kraken-build/src/kraken/std/cargo/tasks/cargo_sync_config_task.py
qsantos: This is confusing because I get the feeling that the behavior changes every day. But from my tests today, and from [Cargo's source code](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/auth/mod.rs#L110), my understanding is that the credential providers should be provided in _reverse_ order of preference.

I have amended the code and added a comment.

kraken-build/src/kraken/std/cargo/tasks/cargo_sync_config_task.py
qsantos: Good point. It's even necessary: since `.cargo/config.toml` is not yet updated when still in the task, `cargo login` fails with `error: authenticated registries require a credential-provider to be configured`. I must have always had a previously-updated `.cargo/config.toml` in my tests.

I have added a task.

kraken-build/src/kraken/std/cargo/tasks/cargo_sync_config_task.py
qsantos: There is not authoritative list of registries in Kraken. Instead, it relies on the contents of the project's `.cargo/config.toml`. That file may contain obsolete registries that do not exist anymore. For some reason, it only happens when there is already a registered token for the registry, so not on the first run.

I have added a comment.

kraken-build/src/kraken/std/cargo/__init__.py
```diff
@@ -182,6 +184,24 @@ def cargo_sync_config(
     return task
 
 
+def cargo_login(
+    *,
+    project: Project | None = None,
+) -> CargoLoginTask:
+    """Creates a task that the :func:`cargo_build` and :func:`cargo_publish`
+    tasks will depend on to login in the Cargo registries"""
```
ntrinquier:
```suggestion
    """Creates a task that will be added to the build and publish support groups
    to login in the Cargo registries"""
```

kraken-build/src/kraken/std/cargo/__init__.py
```diff
@@ -182,6 +184,24 @@ def cargo_sync_config(
...
+
+    project = project or Project.current()
+    cargo = CargoProject.get_or_create(project)
+    task = project.task("cargoLogin", CargoLoginTask, group="apply")
+    task.registries = Supplier.of_callable(lambda: list(cargo.registries.values()))
+    project.group(CARGO_BUILD_SUPPORT_GROUP_NAME).add(task)
```
ntrinquier, line 223:
```suggestion
    project.group(CARGO_BUILD_SUPPORT_GROUP_NAME).add(task)
    project.group(CARGO_PUBLISH_SUPPORT_GROUP_NAME).add(task)
```

daladim:
Notes after a call: we should make sure this MR
* starts the MITM proxy (which is still required, can be disabled later)
* configures cargo to use this proxy
* configures cargo to be aware it is authenticated to a registry (otherwise we would hit a `no token found for 'whatever-registry-here'` error, even if kraken is using the proxy, as it looks kraken does not even attempt a request when it does not know the token)

kraken-build/src/kraken/std/cargo/tasks/cargo_login.py
```diff
@@ -0,0 +1,39 @@
...
+                    # This happens when the project's .cargo/config.toml file
+                    # contains a regitry which does not exist (anymore); since
+                    # that means it is not used, we can just skip configuring
+                    # authentication on this registry
+                    pass
+                else:
+                    # unknown error, fail normally
+                    print(repr(p.stderr))
+                    stderr.write(p.stderr.decode())
+                    p.check_returncode()
```
NiklasRosenstein: I think we're missing a `TaskStatus.error()` here?

qsantos: Is it better with https://github.com/kraken-build/kraken/pull/153/commits/ca364bf58e44eca7f5e1005c7b620ecae8149dd6?

qsantos:
@daladim 
> Notes after a call: we should make sure this MR
> 
>     * starts the MITM proxy (which is still required, can be disabled later)
> 
>     * configures cargo to use this proxy
> 
>     * configures cargo to be aware it is authenticated to a registry (otherwise we would hit a `no token found for 'whatever-registry-here'` error, even if kraken is using the proxy, as it looks kraken does not even attempt a request when it does not know the token)

Regarding the first two points, this MR does not touch `CargoAuthProxyTask`, is that enough?

For the last point, this is what #140 did, by adding `cargo:token` as a dummy `global-credential-providers` to convince Cargo to still attempt the request. This MR only adds more `global-credential-providers`, so it should still work.

NiklasRosenstein:
I've only tested this marginally -- we really need some end-to-end tests that include a sparse and private registry. @Colin1860 recently added a test with https://github.com/d-e-s-o/cargo-http-registry but it's (1) not sparse and (2) accepts and credentials. I'm gonna go ahead with merging this anyway to unblock us integrating it into our internal extension module.

PR #161, opened 2024-01-10
kraken-build/: feature: Add a `PexBuildTask` and add support for most Python tooling tasks to defer to an alternative binary (which the DAG builder can use to point to a PEX)

NiklasRosenstein:
Using separate PEX binaries (or isolated environments) for these Python tools has several advantages:

* We can introduce a dependency on a new tool in our DAG for Python projects without having to get people to update their `pyproject.toml`
* We can control the default version of the tools in the Kraken code
* You can make use of tools that would otherwise have conflicting common dependencies
* The PEX binaries can be cached well in CI or remotely (we don't do individual remote caching for this but we very easily could!)

There's one potential disadvantage for Python projects using Nix however, because these tools would eventually no longer be listed in a Project's `poetry.lock` or `pdm.lock` file. This will require building additional Nix logic that pulls in these dependencies from nixpkgs directly.


kraken-build/pyproject.toml
```diff
@@ -56,6 +56,7 @@ tomli-w = "^1.0.0"
 twine = "^4.0.1"
 typeapi = "^1.4.3"
 typing-extensions = ">=4.6.0"
+pex = "^2.1.156"
```
ntrinquier: Let's keep alphabetical order?

kraken-build/src/kraken/std/python/tasks/flake8_task.py
```diff
@@ -16,6 +16,7 @@ class Flake8Task(EnvironmentAwareDispatchTask):
     description = "Lint Python source files with Flake8."
     python_dependencies = ["flake8"]
 
+    flake8_bin: Property[str] = Property.default("flake8")
```
ntrinquier, line 20: We also need to change the command below to replace `flake8`

kraken-build/src/kraken/std/python/tasks/isort_task.py
```diff
@@ -13,6 +13,7 @@
 class IsortTask(EnvironmentAwareDispatchTask):
     python_dependencies = ["isort"]
 
+    isort_bin: Property[str] = Property.default("isort")
```
ntrinquier, line 17: We also need to change the command below to replace `isort`

kraken-build/src/kraken/std/python/tasks/mypy_task.py
```diff
@@ -21,12 +22,28 @@ class MypyTask(EnvironmentAwareDispatchTask):
...
         #       happens regularly but is hard to detect automatically).
+
         status_file = (self.project.directory / ".dmypy.json").absolute()
-        command = ["dmypy", "--status-file", str(status_file), "run", "--"] if self.use_daemon.get() else ["mypy"]
+        if self.use_daemon.get():
+            command += ["--status-file", str(status_file), "run", "--"]
+        if mypy_pex_bin:
+            # Have mypy pick up the Python executable from the virtual environment that is activated automatically
+            # during the execution of this task as this is an "EnvironmentAwareDispatchTask". If we don't supply this
+            # option, MyPy will only know the packages in it's PEX.
```
ntrinquier:
```suggestion
            # option, MyPy will only know the packages in its PEX.
```

ntrinquier (approved):

PR #171, opened 2024-01-16
Improve Windows support

Robert-Krajewski-Helsing:
* Add support for pyenv-win
* Make subprocess calls for tasks work
* Avoid using dmypy due to non-zero exit code without any error message


kraken-build/src/kraken/std/python/tasks/mypy_task.py
```diff
@@ -10,6 +12,8 @@
 from .base_task import EnvironmentAwareDispatchTask
 
 
+logger = logging.getLogger(__name__)
+
```
NiklasRosenstein:
```suggestion

logger = logging.getLogger(__name__)


```

NiklasRosenstein: 1 line from imports, 2 lines to next definition

kraken-build/src/kraken/common/findpython.py
```diff
@@ -148,6 +148,13 @@ def get_candidates(
             if re.match(r"\d+\.\d+\.\d+$", item.name) and item.is_dir():
                 yield {"path": str(item / "bin" / "python"), "exact_version": item.name}
 
+    # pyenv (Windows)
+    pyenv_versions = (Path(os.getenv('PYENV')) / "versions").expanduser()
```
NiklasRosenstein:
```suggestion
    pyenv_versions = (Path(os.getenv("PYENV")) / "versions").expanduser()
```

Black will complain about this

NiklasRosenstein (commented):
Thanks @Robert-Krajewski-Helsing ! Left a few comments. If you could also give this a changelog entry, this seems good to go!

PR #182, opened 2024-01-31
feat: add flags to clippy to cover more cases

tomkarw:
* `--workspace`, tests whole workspace, important as for non-virtual workspace clippy will lint root crate only
* `--all-features`, tests with all features enabled, important to prevent committing broken code behind a feature flag
* `-- -D warnings`, fail on warnings, should be the default behavior in pipelines

After internal discussions I opted against setting `deny_warnings` to `True` by default, so no "breaking changes" here.  Defaulting to `True` for `--workspace` and `--all-features` thus also won't break anything, and should be set by default, as its absence is basically a foot-gun.

### DIFF
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -1,5 +1,6 @@
 [[entries]]
 id = "c735fee0-cd87-47a6-8c6f-b52463f0f73e"
 type = "feature"
-description = "Add flags to CargoClippyTask to cover more cases"
+description = "Add `workspace`, `all_features` and `deny_warnings` properties to `CargoClippyTask`."
 author = "@tomkarw"
+pr = "https://github.com/kraken-build/kraken/pull/182"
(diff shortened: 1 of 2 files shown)


.changelog/_unreleased.toml
```diff
@@ -0,0 +1,5 @@
+[[entries]]
+id = "c735fee0-cd87-47a6-8c6f-b52463f0f73e"
+type = "feature"
+description = "Add flags to CargoClippyTask to cover more cases"
```
```suggestion
description = "Add `workspace`, `all_features` and `deny_warnings` properties to `CargoClippyTask`."
```

.changelog/_unreleased.toml
```diff
@@ -0,0 +1,5 @@
...
+author = "@tomkarw"
```
NiklasRosenstein, line 5:
```suggestion
author = "@tomkarw"
pr = "https://github.com/kraken-build/kraken/pull/182"
```

Unfortunately the automation to inject this field is broken atm. 😄

NiklasRosenstein (approved):
LGTM!

PR #195, opened 2024-02-07
`pytest`: Append to coverage report

antoinebrl:
`pytest` is used twice, once for unit tests and once for integration tests. The second execution no longer overrides the coverage report generated by the first one.


NiklasRosenstein:
Nice, I did not realize that was an option. 😃 The only thing we should probably make sure is that we don't continue to append and append to the same coverage report -- i.e. the first `pytest` run should clear the coverage report.

PR #196, opened 2024-02-09
Move version bump logic into publish task

Colin1860:
Since bumping versions is only really necessary when publishing the affected logic is moved into the publish task and the dedicated bump-task removed


kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -53,3 +65,49 @@ def make_safe(self, args: list[str], env: dict[str, str]) -> None:
...
+    def execute(self) -> TaskStatus:
+        version = self.version.get()
+        if version is not None:
+            content = self._get_updated_cargo_toml(version)
+            with atomic_file_swap(self.cargo_toml_file.get(), "w", always_revert=True) as ctx:
+                ctx.write(content)
+                ctx.close()
+                return super().execute()
+        else:
+            return super().execute()
```
NiklasRosenstein:
```suggestion
        with contextlib.ExitStack() as stack:
            if (version := self.version.get()) is not None:
                fp = stack.enter_context(atomic_file_swap(self.cargo_toml_file.get(), "w", always_revert=True))
                fp.write(content)
                fp.close()
            return super().execute()
```

I think something like this is more concise

NiklasRosenstein (commented):
Looks good to me overall! :)

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
Colin1860: I've had something similar like this but then, python complains that all branches of execution must return a value, and if i move the return statement outside of the exitstack scope, the file is reverted before super().execute()

NiklasRosenstein: Ah, yes this is a bit of an unfortunate quirk with the `with` statement and type checking imo. Using

```py
    result = super().execute()
return result
```

should work?

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -53,3 +65,48 @@ def make_safe(self, args: list[str], env: dict[str, str]) -> None:
...
+                    dependency["registry"] = registry_alias
+
+    def execute(self) -> TaskStatus:
+        with contextlib.ExitStack() as stack:
+            if (version := self.version.get()) is not None:
+                content = self._get_updated_cargo_toml(version)
+                fp = stack.enter_context(atomic_file_swap(self.cargo_toml_file.get(), "w", always_revert=True))
+                fp.write(content)
+                fp.close()
+                result = super().execute()
```
NiklasRosenstein: This has one too many indent. We want to run super.execute regardless of whether a version is set to bump or not.

NiklasRosenstein (changes requested):

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
Colin1860: Alright, that's what i thought. But then i won't get around writing `super().execute()` twice i think

Colin1860: Or this?
```
result = TaskStatus.pending()
with contextlib.ExitStack() as stack:
    if (version := self.version.get()) is not None:
        content = self._get_updated_cargo_toml(version)
        fp = stack.enter_context(atomic_file_swap(self.cargo_toml_file.get(), "w", always_revert=True))
        fp.write(content)
        fp.close()
    result = super().execute()

return result
```

Colin1860: Ok, now i understand. Didn't know this was possible without first initializing the result variable.

PR #197, opened 2024-02-12
`None` checks on buffrs/manifest

vsiles:
Importing some changes suggested by @alexespencer. These would catch NPEs and spurious updating of the dict with `None`s

### DIFF
--- a/kraken-build/src/kraken/std/buffrs/manifest.py
+++ b/kraken-build/src/kraken/std/buffrs/manifest.py
@@ -27,10 +27,11 @@ class BuffrsManifest:
 
     @classmethod
     def of(cls, path: Path, data: dict[str, Any]) -> BuffrsManifest:
+        package = data.get("package")
         return cls(
             path,
             data,
-            Package.from_json(data["package"]) if "package" in data else None,
+            package=Package.from_json(package) if package is not None else None,
         )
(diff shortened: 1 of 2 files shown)


kraken-build/src/kraken/std/buffrs/manifest.py
```diff
@@ -27,10 +27,11 @@ def read(cls, path: Path) -> BuffrsManifest:
 
     @classmethod
     def of(cls, path: Path, data: dict[str, Any]) -> BuffrsManifest:
+        package = data.get("package")
```
alexespencer, line 30:
```suggestion
        package = data.get("package", None)
```
Might this be needed in case "package" isn't a key?

vsiles: I think we are good: from [this](https://docs.python.org/3/library/stdtypes.html#dict):

> Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError).

This would have happened if we had used `data["package"]` which can raise `KeyError`

alexespencer (approved):

qsantos (approved):

PR #202, opened 2024-02-20
Make Rust builds respect Cargo.lock when present

jonhoo:
Kraken should, by default, respect the `Cargo.lock` present in the current Cargo project rather than silently pull down dependencies not explicitly noted in the initial `Cargo.lock`. If the lockfile is out of date, the underlying package should be updated to reflect that change if `Cargo.lock` is checked in.

To avoid breaking existing workflows that don't check in `Cargo.lock`, the default behavior is to only pass `--locked` if a `Cargo.lock` exists in the parent directory tree of the current working directory.

The flag only really matters to CargoBuildTask, CargoTestTask, and CargoManifest, but is also propagated to CargoPublishTask so that it isn't allowed to silently update the lockfile if it happens to be called first. Clippy doesn't take `--locked`, and so remains unmodified.

Note also that checking in `Cargo.lock` is now recommended by the Rust project: https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

### DIFF
--- a/kraken-build/src/kraken/std/cargo/manifest.py
+++ b/kraken-build/src/kraken/std/cargo/manifest.py
@@ -80,7 +80,8 @@ class CargoMetadata:
             str(project_dir / "Cargo.toml"),
         ]
         if locked is None:
-            for parent in (project_dir / "Cargo.toml").absolute().parents:
+            project_dir = project_dir.absolute()
+            for parent in [project_dir, *project_dir.parents]:
                 if (parent / "Cargo.lock").exists():
                     cmd.append("--locked")
                     break
--- a/kraken-build/src/kraken/std/cargo/manifest.py
+++ b/kraken-build/src/kraken/std/cargo/manifest.py
@@ -70,7 +70,7 @@ class CargoMetadata:
     target_directory: Path
 
     @classmethod
-    def read(cls, project_dir: Path) -> CargoMetadata:
+    def read(cls, project_dir: Path, locked: bool | None = None) -> CargoMetadata:
         cmd = [
             "cargo",
             "metadata",
@@ -79,6 +79,15 @@ class CargoMetadata:
             "--manifest-path",
             str(project_dir / "Cargo.toml"),
         ]
+        if locked is None:
+            for parent in (project_dir / "Cargo.toml").absolute().parents:
+                if (parent / "Cargo.lock").exists():
+                    cmd.append("--locked")
+                    break
+        elif locked:
+            # if locked is True, we should *always* pass --locked.
+            # the expectation is that the command will fail w/o Cargo.lock.
+            cmd.append("--locked")
         result = subprocess.run(cmd, stdout=subprocess.PIPE)
         if result.returncode != 0:
             logger.error("Stderr: %s", result.stderr)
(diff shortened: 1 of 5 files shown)


kraken-build/src/kraken/std/cargo/manifest.py
```diff
@@ -79,6 +79,15 @@ def read(cls, project_dir: Path) -> CargoMetadata:
             "--manifest-path",
             str(project_dir / "Cargo.toml"),
         ]
+        if locked is None:
+            for parent in (project_dir / "Cargo.toml").absolute().parents:
```
NiklasRosenstein: I wonder if we should stop at some kind of boundary (max depth, not go beyond the home directory)? Having a rogue `Cargo.toml` lying around could mess this up. (This has happened before; someone had a rogue `pyproject.toml` lying in the home directory that altered the behavior of Poetry (iirc) and it took a while to debug).

jonhoo: Yeah, I wanted something like that too, but couldn't really think of any stop heuristics that don't bring their own problems. Stopping at the home directory gets trick cross-platform. We _can_ do it, but would need either a library or heuristics to determine what `$HOME` is. Stopping when we no longer see a `Cargo.toml` is tempting, but would break for workspaces where the members are two levels deep in the hierarchy rather than one without a `Cargo.toml` in between (which is possible). A heuristic like "stop if you've seen two directories without `Cargo.toml`s after seeing at least one `Cargo.tom`, but that gets a) pretty arbitrary, and b) incurs the risk of picking up strays again.

One thing worth highlighting here is that we're looking for `Cargo.lock`, _not_ `Cargo.toml`. And my suspicion is that stray `Cargo.lock`s are much less common than stray `Cargo.toml`s.

NiklasRosenstein: Yep, I buy those arguments!

One last question: The join with `Cargo.toml` is only to make sure we also check `project_dir`, yes? In that case, I think this is less distracting:

```suggestion
            for parent in (project_dir, *project_dir.parents):
```

jonhoo: Yep, that's why, although I think we still need `.absolute()`, because otherwise `.parents` would be empty when `project_dir` is `.` for example. At least my understand of `parents` is that it operates on `dir.split("/")` rather than resolving `..` at each step (but I could be wrong about that).

Also, wouldn't `()` need to become `[]`?

NiklasRosenstein: Agreed, we should keep it. I accidentally dropped it from the suggestion!

`()` means you'll make a tuple instead of a list; though some simple testing I just did with `timeit` indicates that `[]` is actually faster (obviously this is a negligible optimization here but it's effectively the only difference that could be relevant here).

jonhoo: Is
```python
for parent in [project_dir, *project_dir.absolute().parents]:
```
clearer than what's there now though?

NiklasRosenstein: I think the intent is clearer, yes. Could move the `.absolute()` to a separate line to make it more readable as well

NiklasRosenstein:
```suggestion
            project_dir = project_dir.absolute()
            for parent in [project_dir, *project_dir.parents]:
```

PR #216, opened 2024-03-11
fix: strip trailing whitespaces from gitignore.io responses

### LINKED ISSUE #81: Trim trailing whitespace for gitignore tokens from gitignore.io
Sometimes you may end up with this diff after running `krakenw apply` because you manually added an entry to `.gitignore` and your editor stripped trailing whitespace on lines in the file. 

image

tomkarw:
gitignore.io sometimes returns files with trailing whitespaces at the end of lines. This causes issues when editors remove the whitespaces on formatting, causing autogenerated parts of .gitignore file to change and fail kraken check.

Closes https://github.com/kraken-build/kraken/issues/81.


NiklasRosenstein:
As discussed offline, the issue with the `Icon` line is that it's intentionally followed by a carriage-return. But we should probably just consider throwing out the `gitignore.io` stuff again and go with a simpler set of well-defined Gitignore lines instead.

NiklasRosenstein:
Hey @tomkarw 👋 Would you still like to work on this, and update this MR and replace the `gitignore.io` stuff with a curated list of Gitignore lines?

PR #219, opened 2024-03-14
improvement: Remove mitmweb dependency, install it with PEX instead. Print mitmweb proxy and UI URLs in the CargoAuthProxy task

NiklasRosenstein:


NiklasRosenstein:
CI failures are sporadic

PR #221, opened 2024-03-15
fix: Fix that `pyproject.toml` is continuously outdated in certain cases

NiklasRosenstein:
Sometimes the `pyproject.toml` would add or remove some newlines, in certain cases also be continuously outdated, never satisfying the `pyproject.check` task. This is because we converted the top-level `tomlkit` container that keeps track of formatting to a standard Python `dict`.


NiklasRosenstein:
Now that we're using a `MutableMapping` as the base class for `Pyproject` since #223, `setdefault()` behaves differently as explained in https://github.com/sdispater/tomlkit/issues/49#issuecomment-1999713939.

Need to override `setdefault()`, although I fear we might hit other incompatibilities issues down the road.. (This is the second, after #223).

PR #224, opened 2024-03-15
Use `pytest-xdist` and `PYTEST_FLAGS=--numprocesses auto` in CI

NiklasRosenstein:


NiklasRosenstein:
CI failures are due to sporadic issues (Dockerhub rate limit and the uv-installer test seems broken atm.)

PR #236, opened 2024-04-11
cleanup `pyproject.toml`

NiklasRosenstein:


NiklasRosenstein:
We don't actually embed these defaults in Kraken, yet.

PR #238, opened 2024-04-29
fix: inject version for path dependencies in workspace virtual manifests

tomkarw:
There was an unhandled case in version injection task.

For a project with virtual manifest like this:
```
# in /Cargo.toml

[workspace.dependencies]
subcrate = { path = "./subcrate" }
```
Kraken would not inject `version` and `registry` attributes.

This MR fixes it, by correctly injecting versions for dependencies declared in workspaces with virtual manifest.

Decisions:

- Add `_raw` attribute to most manifest related classes, store raw JSON data there (without removing the handled fields)
- For fields that already had a subset of this, rename the field to `_raw` and use a getter with deprecation warning for backward compatibility
- All `to_json` methods could be simplified, as they can simply return `self._raw`

### DIFF
--- a/kraken-build/src/kraken/std/cargo/manifest.py
+++ b/kraken-build/src/kraken/std/cargo/manifest.py
@@ -6,13 +6,14 @@ import json
 import logging
 import os
 import subprocess
-from dataclasses import dataclass, fields
+from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
 from typing import Any
 
 import tomli
 import tomli_w
+from deprecated import deprecated
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


kraken-build/src/kraken/std/cargo/manifest.py
```diff
@@ -161,16 +161,16 @@ def from_json(cls, json: dict[str, str]) -> Package:
...
 
     def to_json(self) -> dict[str, str]:
-        values = {f.name: getattr(self, f.name) for f in fields(self) if f.name != "unhandled"}
-        if self.unhandled is not None:
-            values.update({k: v for k, v in self.unhandled.items() if v is not None})
-        return {k: v for k, v in values.items() if v is not None}
+        return {k: v for k, v in self.raw if v is not None}
 
+    @property
+    def unhandled(self) -> dict[str, Any]:
```
NiklasRosenstein, line 179:
```suggestion
    @deprecated(reason="use Package.raw instead", version="0.37.0")
    def unhandled(self) -> dict[str, Any]:
```

kraken-build/src/kraken/std/cargo/manifest.py
```diff
@@ -179,36 +179,43 @@ def from_json(cls, json: dict[str, str]) -> WorkspacePackage:
...
```
NiklasRosenstein, line 199:
```suggestion
    @deprecated(reason="use WorkspacePackage.raw instead", version="0.37.0")
    def unhandled(self) -> dict[str, Any]:
```

kraken-build/src/kraken/std/cargo/manifest.py
```diff
@@ -179,36 +179,43 @@ def from_json(cls, json: dict[str, str]) -> WorkspacePackage:
...
             "package": self.package.to_json() if self.package else None,
             "members": self.members if self.members else None,
-        }
-        if self.unhandled is not None:
-            values.update({k: v for k, v in self.unhandled.items() if v is not None})
+        })
         return {k: v for k, v in values.items() if v is not None}
 
+    @property
+    def unhandled(self) -> dict[str, Any]:
```
NiklasRosenstein, line 226:
```suggestion
    @deprecated(reason="use Workspace.raw instead", version="0.37.0")
    def unhandled(self) -> dict[str, Any]:
```

NiklasRosenstein (commented):
LGTM overall; would be nice to have a test that checks if the version gets injected correctly into the workspace dependencies!

PR #239, opened 2024-05-14
Make it possible for Kraken to handle PyPI

qsantos:
Explicitly listing PyPI as a package index with an URL results in the error below. This PR sets the `index_url` of the package indexes whose aliases matches `pypi` to `None`. This requires another change to handle this `None` value properly.

```
The PyPI repository cannot be configured with a custom url.
```

### DIFF
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -1,0 +1,5 @@
+[[entries]]
+id = "e6f6ee7b-e382-44f1-84b8-a24ad01c5d24"
+type = "hygiene"
+description = "Make it possible for Kraken to handle PyPI"
+author = "quentin.santos@helsing.ai"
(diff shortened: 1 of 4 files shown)


NiklasRosenstein (changes requested):
Hi Quentin, thanks for the PR!

Instead of updating the Kraken API to "feel" more like the Poetry API, I think I'd rather that we update the `PoetryBuildSystem` to handle the Python package indices from the Kraken Python-project API correctly.

I.e., according to your changes it appears to me that Poetry might not accept explicitly setting the `index_url` when the source is `PyPI`; but instead of Kraken now making the URL optional, the `PoetryBuildSystem` implementation can skip setting that field in `pyproject.toml`.

This also means we don't need to accept alternative case aliases for `pypi` (i.e. no need to call `alias.lower()` here).

PR #250, opened 2024-06-13
improvement: Replace `rich` logger with `loguru` which produces nicer log formatting and requires less setup

NiklasRosenstein:
Kind of a bit blocked on https://github.com/Delgan/loguru/issues/1151, since we use printf formatting quite extensively.


NiklasRosenstein:
Seems we might also be facing an issue with `dill` pickling.

PR #256, opened 2024-07-02
Update manifest.py

frk1993:
Allow reading cargo.toml configuration from private CargoManifest._data. This will increase the flexibility of `CargoManifest`. For example if a package has conditional target dependencies, currently we cannot access them.


NiklasRosenstein:
Doesn't seem like we needed this yet. Closing for now; happy to accept this though but I think we can just rename `_data` to `data` or expose it as a property.

PR #280, opened 2024-08-16
Cargo - Skip existing packages when publishing

jcaille:
Flaky CI can cause issues where some packages have been published, and others have not, causing re-runs of the CI to fail.

This PR introduces a new `prepare()` step in the CargoPublishTask, which verifies the existence of a package in the remote registry, and skips the task entirely if the package already exists at the given version. This relies on reading from the Cargo Registry's [Index](https://doc.rust-lang.org/cargo/reference/registry-index.html).

A new `allow_overwrite` option is introduced to skip this verification and always try publishing the package.


kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -37,6 +41,36 @@ class CargoPublishTask(CargoBuildTask):
...
 
+    #: Allow Overwrite of existing packages
+    allow_overwrite: Property[bool] = Property.default(False)
+
+    def prepare(self) -> TaskStatus | None:
+        """Checks if the crate@version already exists in the registry. If so, the task will be skipped"""
+        if self.allow_overwrite.get():
+            return TaskStatus.pending()
+
+        package_name = self.package_name.get()
```
NiklasRosenstein: I'm not 100% sure on this, but I _believe_ this may be the local workspace member name instead of the full `name` from the `Cargo.toml`!

NiklasRosenstein: Tbh. I'm also not sure why we have it already as a property when it's not referenced in the implementation of `execute()` before.. 🤔

jcaille: [cargo_publish_task.py#L86-L88](https://github.com/kraken-build/kraken/blob/86d17ed/kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py#L86-L88) hints that it can be None. I guess in that case, we should pick the package_name from the `cargo.toml` file

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -108,3 +144,77 @@ def execute(self) -> TaskStatus:
                 fp.close()
             result = super().execute()
         return result
+
+    @staticmethod
+    def _sanitize_version(version: str) -> str:
+        """
+        Cargo does not play nicely with semver metadata (ie. 1.0.1-dev3+abc123)
+         We replace that to 1.0.1-dev3abc123
```
NiklasRosenstein:
```suggestion
        We replace that to 1.0.1-dev3abc123
```

nit

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
NiklasRosenstein: After digging a bit deeper into the usage of `package_name` in our downstream internal Kraken module I found:

```py
        for member in self.metadata.workspaceMembers:
            if member.name not in self.exclude:
                members_publish_task[member.name] = cargo_publish(
                    publish_repo,
                    incremental=False,
                    verify=not is_playground and self.verify,
                    retry_attempts=5,
                    package_name=member.name,
                    project=self.project,
                    version=version,
                    cargo_toml_file=member.manifest_path,
                )
```

Where `metadata` is a `CargoMetadata` which is generated with `CargoMetadata.read()`, and ends up deriving the `member.name` here:

https://github.com/kraken-build/kraken/blob/62cd7f859d350a8c5616e223c460ad66afb0acd5/kraken-build/src/kraken/std/cargo/manifest.py#L127-L133

So I think we're good here, `package_name` is populated with the actual Crate name and not the [workspace member name](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields).

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -84,14 +115,19 @@ def _get_updated_cargo_toml(self, version: str) -> str:
             if manifest.dependencies:
                 self._push_version_to_path_deps(fixed_version_string, manifest.dependencies.data, registry.alias)
             if manifest.build_dependencies:
-                self._push_version_to_path_deps(fixed_version_string, manifest.build_dependencies.data, registry.alias)
+                self._push_version_to_path_deps(
+                    fixed_version_string,
+                    manifest.build_dependencies.data,
+                    registry.alias,
```
NiklasRosenstein: nit: unrelated reformatting?

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -84,14 +115,19 @@ def _get_updated_cargo_toml(self, version: str) -> str:
...
+                )
         return manifest.to_toml_string()
 
     def _push_version_to_path_deps(
         self, version_string: str, dependencies: dict[str, Any], registry_alias: str
     ) -> None:
         """For each dependency in the given dependencies, if the dependency is a `path` dependency, injects the current
-        version and registry (required for publishing - path dependencies cannot be published alone)."""
+        version and registry (required for publishing - path dependencies cannot be published alone).
+        """
```
NiklasRosenstein: nit: unrelated reformatting?

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -108,3 +144,77 @@ def execute(self) -> TaskStatus:
...
+                )
+            session.auth = registry.read_credentials
+            config_response = session.get(f"{index}/config.json")
+            if config_response.status_code % 200 != 0:
+                logger.warn(config_response.text)
+                return TaskStatus.pending(
+                    "Unable to verify package existence - failed to download config.json file from registry"
+                )
+
+        # >> Index files layout
```
NiklasRosenstein, line 182:
```suggestion
        # >> Index files layout
        # Reference: https://doc.rust-lang.org/cargo/reference/registry-index.html#index-files
```

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -108,3 +144,77 @@ def execute(self) -> TaskStatus:
...
+            path = ["2"]
+        elif len(package_name) == 3:
+            path = ["3", package_name.lower()[0]]
+        else:
+            package_name_lower = package_name.lower()
+            path = [package_name_lower[0:2], package_name_lower[2:4]]
+
+        # >> Download the package file
+        package_path = "/".join(path + [package_name])
+        package_response = session.get(f"{index}/{package_path}")
```
NiklasRosenstein:
```suggestion
        # >> Download the index file
        index_path = "/".join(path + [package_name])
        index_response = session.get(f"{index}/{package_path}")
```

nit

kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
```diff
@@ -77,6 +77,26 @@ def publish_lib_and_build_app(repository: CargoRepositoryWithAuth, tempdir: Path
...
+            cargo_auth_proxy()
+            task = cargo_sync_config()
+            task.git_fetch_with_cli.set(True)
+            cargo_check_toolchain_version(minimal_version="1.60")
+            cargo_publish(
+                cargo_registry_id,
+                version=publish_version,
+                cargo_toml_file=project1.directory.joinpath("Cargo.toml"),
+            )
+            project1.context.execute(["fmt", "lint", "publish"])
```
NiklasRosenstein: We should assert that the task status indicates that it was skipped. One way to do this would be to make `Context.execute()` return the `TaskGraph`, and then use `TaskGraph.get_status(cargo_publish_task)`.

NiklasRosenstein (changes requested):
Thanks a lot @jcaille !

Nits are not blocking, but I think the assertion for the Task being skipped is important.

kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
jcaille: Will do - also noticed that this should be using `project2`. Oops

jcaille:
PR is mostly ready.

We unfortunately can't test as-is in the integration setup, as the private registry we use in the testing harness does not support the sparse protocol. We could "hack" around simply accessing `files://` if you feel strongly about it.

The downside of not testing is limited, since in the worse case scenario, we keep the old behavior of "always publish everything"

jcaille:
After chatting with Niklas offline :

- The current tests were not verifying the new task behavior, since [cargo-http-registry](https://github.com/d-e-s-o/cargo-http-registry#:~:text=cargo-http-registry) does not support the sparse protocol
- As an alternative, we could be using [cratery](https://github.com/cenotelie/cratery#:~:text=cratery), but this crate seems a bit more involved (e.g. requires seem to OAuth setup)
- So I mocked a server to verify that the task was skipped if the package already exists

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -108,3 +138,79 @@ def execute(self) -> TaskStatus:
...
+        else:
+            package_name_lower = package_name.lower()
+            path = [package_name_lower[0:2], package_name_lower[2:4]]
+
+        # >> Download the index file
+        index_path = "/".join(path + [package_name])
+        index_response = session.get(f"{index}/{index_path}")
+
+        if index_response.status_code in [404, 410, 451]:
+            return TaskStatus.pending("Package {package_name} does not already exists in {registry.alias}")
```
NiklasRosenstein:
```suggestion
            return TaskStatus.pending(f"Package {package_name} does not already exists in {registry.alias}")
```

kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py
```diff
@@ -37,6 +42,35 @@ class CargoPublishTask(CargoBuildTask):
...
+        if not package_name:
+            return TaskStatus.pending("Unable to verify package existence - unknown package name")
+        if not version:
+            return TaskStatus.pending("Unable to verify package existence - unknown version")
+
+        try:
+            return self._check_package_existence(package_name, version, self.registry.get())
+        except Exception as e:
+            logger.warn(
+                f"An error happened while checking for {package_name} existence in {self.registry.get().alias}, {e}",
```
NiklasRosenstein:
```suggestion
                "An error happened while checking for {} existence in %s, %s",
                package_name,
                self.registry.get().alias,
                e,
```

Formatting the log message directly is discouraged as it evaluates eagerly and can lead to substitution errors if the resulting message string contains placeholders.

NiklasRosenstein (approved):

kraken-build/src/kraken/std/cargo/__init__.py
```diff
@@ -427,6 +427,7 @@ def cargo_publish(
     project: Project | None = None,
     version: str | None = None,
     cargo_toml_file: Path = Path("Cargo.toml"),
+    allow_overwrite: bool = False,
```
NiklasRosenstein, line 430: FYI because this changes semantics of the task, I will release this as a minor update.

PR #281, opened 2024-08-19
improvement: Use UV as the default installer for Kraken build-time virtual environments

NiklasRosenstein:
This MR switches to [UV](https://pypi.org/project/uv/) as the default installer for Kraken-wrapper runtime virtual environments. UV-support in Kraken has been in testing for a while and has shown great promise, reducing installation times of virtual environments by at least one magnitude.

We depend on the `uv` package directly already so this should just work out of the box for everyone that picks up the latest `kraken-wrapper` version. Existing virtual environments will continue to use the same installer as the information on the installer used is written into the virtual environment and re-used on upgrades/reinstalls.


Tpt (approved):
Would be amazing to get shipped.

NiklasRosenstein:
I believe @ulucs had some concerns for users installing `kraken-wrapper` via Nix, where the `uv` binary cannot be determined by `uv.find_uv_bin()` function. 

Otherwise, I don't think anything stands against merging this.

ulucs:
We should be able to fix the issue in `uv` with a patch to its python module; so no more concerns from me on this point

PR #284, opened 2024-08-26
Extend pytest task to accept sequence

sebimarkgraf:


kraken-build/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -46,14 +46,19 @@ def is_skippable(self) -> bool:
...
+        if not tests_dir or len(tests_dir) == 0:
             print("error: no test directory configured and none could be detected")
             return TaskStatus.failed("no test directory configured and none could be detected")
+
         command = [
             "pytest",
             "-vv",
             str(self.project.directory / self.settings.source_directory),
-            str(self.project.directory / tests_dir),
+            *map(lambda t: str(self.project.directory / t), tests_dir),
```
NiklasRosenstein: nit: consistency with next line (change both or keep same methodology)

NiklasRosenstein (commented):
LGTM, but you need to fix Mypy and formatting and I got one nit. 🙂

kraken-build/src/kraken/std/python/tasks/pytest_task.py
sebimarkgraf: Went with the pre-existing style

.changelog/_unreleased.toml
```diff
@@ -0,0 +1,5 @@
+[[entries]]
+id = "e3dd07f8-7520-45b9-9407-9f650e4d333c"
+type = "feature"
+description = "Extend pytest task to accept sequence"
```
```suggestion
description = "Extend pytest task to accept a sequence of paths that contain test files."
```

kraken-build/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -46,14 +46,19 @@ def is_skippable(self) -> bool:
...
             print("error: no test directory configured and none could be detected")
             return TaskStatus.failed("no test directory configured and none could be detected")
+
         command = [
             "pytest",
             "-vv",
             str(self.project.directory / self.settings.source_directory),
-            str(self.project.directory / tests_dir),
+            *[str(self.project.directory / path) for path in tests_dir],
             *[str(self.project.directory / path) for path in self.include_dirs.get()],
```
NiklasRosenstein, line 69: Hold on, I only see this now. This is already the property for passing the `[2...n]`th path.

Though the API seems a bit ~what~ whack if you have to pick one main "tests_dir" and then any number of "include_dirs", so I think we should deprecate this one.

Can you add a `warnings.warn("...", DeprecationWarning)` call
if include_dirs is set?

kraken-build/tests/kraken_std/python/tasks/test_pytest_task.py
```diff
@@ -0,0 +1,42 @@
+from kraken.core import Project
+
+# Renamed import to prevent name collision
```
NiklasRosenstein: Seems a bit redundant. :)

kraken-build/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -30,7 +31,7 @@ class PytestTask(EnvironmentAwareDispatchTask):
     description = "Run unit tests using Pytest."
     python_dependencies = ["pytest"]
 
-    tests_dir: Property[Path]
+    tests_dir: Property[Sequence[Path] | Path]
     include_dirs: Property[Sequence[Path]] = Property.default(())
```
NiklasRosenstein:
```suggestion
    include_dirs: Property[Sequence[Path]] = Property.default(())  # NOTE: Deprecated in favour of `tests_dir`
```

NiklasRosenstein (approved):
🚀

PR #290, opened 2024-10-04
Allows to use Maturin with UV

Tpt:


kraken-build/src/kraken/std/python/buildsystem/__init__.py
```diff
@@ -174,14 +174,18 @@ def detect_build_system(project_directory: Path) -> PythonBuildSystem | None:
...
+        if "[tool.poetry" in pyproject_content:
             from .maturin import MaturinPoetryPythonBuildSystem
 
             return MaturinPoetryPythonBuildSystem(project_directory)
-        else:
+        elif "[tool.pdm" in pyproject_content:
             from .maturin import MaturinPdmPythonBuildSystem
 
             return MaturinPdmPythonBuildSystem(project_directory)
+        else:
```
vsiles, line 185: Do we want to avoid the fallback and test `elif "[tool.uv"`too ? Raising an error if no known system matches ?
That would be a better UX in the end to see "build system not supported" rather than defaulting to one of them silently.

vsiles (approved):
@NiklasRosenstein in case you're around sometimes, this would be a really nice to have to start migrating some projects to uv ;)

kraken-build/src/kraken/std/python/buildsystem/__init__.py
Tpt: I am a bit reluctant to remove the fallback that is nice to get a default working for new projects. However, I have added a warning to make it easier for the user to know what is happening. WDYT?

vsiles: Better ;) Good enough for me, thank you

kraken-build/src/kraken/std/python/buildsystem/__init__.py
```diff
@@ -174,22 +174,31 @@ def detect_build_system(project_directory: Path) -> PythonBuildSystem | None:
...
+            from .maturin import MaturinUvPythonBuildSystem
+
+            if "[tool.uv]" not in pyproject_content:
```
NiklasRosenstein:
```suggestion
            if "[tool.uv" not in pyproject_content:
```

Just like the checks for poetry/pdm?

kraken-build/src/kraken/std/python/buildsystem/__init__.py
```diff
@@ -174,22 +174,31 @@ def detect_build_system(project_directory: Path) -> PythonBuildSystem | None:
...
+                    project_directory,
+                )
+            return MaturinUvPythonBuildSystem(project_directory)
 
     if "pdm" in pyproject_content:
         from .pdm import PDMPythonBuildSystem
 
         return PDMPythonBuildSystem(project_directory)
 
     if "[tool.uv]" not in pyproject_content:
```
NiklasRosenstein:
```suggestion
    if "[tool.uv" not in pyproject_content:
```

For consistency with the above?

NiklasRosenstein (approved):
This looks good to me, left a comment.

PR #294, opened 2025-01-07
Release 0.40.5

ghadjigeorghiou:


NiklasRosenstein:
The `rust-poetry-project` test fails, presumably because something changed since the Poetry 2.0.0 release (from 2 days ago) that made it break.

Tpt:
CI fix in #295

PR #297, opened 2025-01-22
Support management of package indexes for uv

antoinebrl:
When `uv` was integrated it was likely still in its infancy and its main purpose was to be a fast inplace replacement for `pip`. Since, `uv` has grown into a full package and project manager. This PR rework the `uv` integration to leverage the latest feature. The initial focus is on the management of packages indexes.


kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -39,14 +37,17 @@ def find_uv_bin() -> str: ...
...
 Safe = Annotated[T, "safe"]
 Unsafe = Annotated[T, "unsafe"]
 
 
 @dataclass
-class PipIndex:
+class UvIndex:
+    # https://docs.astral.sh/uv/configuration/indexes/#defining-an-index
     url: str
+    name: str = ""
```
vsiles: `str | None` ? or do we only encode the named indexes here (and then should avoid an empty string default value ?)

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -62,136 +63,143 @@ def unsafe_url(self) -> str:
...
         return env
 
 
 class UvPyprojectHandler(PyprojectHandler):
     """Implements the PyprojectHandler interface for UV projects."""
 
-    # TODO: Support `uv.toml` configuration file?
-
-    # PyprojectHandler
+    # TODO: Support global `uv.toml` configuration file?
```
vsiles, line 143: I'd avoid something as system dependent as "global", but supporting an "external" `uv.toml` provided by its path might be nice.

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -62,136 +63,143 @@ def unsafe_url(self) -> str:
...
-            hostname = urlparse(url).hostname
-            assert hostname is not None, "expected hostname in package index URL"
-            return f"hostname-{md5(url.encode()).hexdigest()[:5]}"
-
         indexes: list[PackageIndex] = []
-        config: dict[str, Any] = self.raw.get("tool", {}).get("uv", {})
+        for index in self.raw.get("tool", {}).get("uv", {}).get("index", []):
+            indexes.append(
+                PackageIndex(
+                    alias=index.get("name", ""),
```
vsiles, line 159: we might want to us a special string instead of `""` to make debugging easier

vsiles: Or `None` :p

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -62,136 +63,143 @@ def unsafe_url(self) -> str:
...
+                    verify_ssl=True,
+                )
+            )
 
-        if index_url := config.get("index-url"):
+        if index_url := self.raw.get("tool", {}).get("uv", {}).get("index-url"):
             indexes.append(
                 PackageIndex(
-                    alias=gen_alias(index_url),
+                    alias="",  # unnamed index
```
vsiles, line 173: Might want an optional type for alias instead of that

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -244,9 +254,9 @@ def build_v2(self, settings: PythonSettings, output_directory: Path) -> list[Pat
             if shutil.which("uv") != self.uv_bin:
                 env["PATH"] = str(Path(self.uv_bin).parent) + os.pathsep + env["PATH"]
 
-            # We can't pass the --index-url and --extra-index-url options to UV via the pyproject-build CLI,
+            # We can't pass the --default-index and --index options to UV via the pyproject-build CLI,
```
vsiles, line 261: Not sure what this comment is about. Is it up to date ?

kraken-build/src/kraken/std/python/buildsystem/uv.py
antoinebrl: I agree!

kraken-build/src/kraken/std/python/buildsystem/uv.py
antoinebrl: While other package managers like `pdm`, `poetry` and `slap` require that their indexes are named, `uv` doesn't. Hence, `alias` is marked as a required string. I originally wanted to represent the the absence of the index name with `None`, but this had cascading effects on all the call sites for mypy. Indeed, other build systems or environment managers would have to ensure `alias` is not known every time they consume that value. Although maybe sub-optimal, I decided to limit the depth of the changes and use an empty string.

kraken-build/src/kraken/std/python/buildsystem/uv.py
antoinebrl: see comment above: https://github.com/kraken-build/kraken/pull/297#discussion_r1925342139

kraken-build/.changelog/_unreleased.toml
```diff
@@ -0,0 +1,5 @@
+[[entries]]
+id = "faec54d3-e41a-414e-ac47-e4208cf6df2e"
+type = "improvement"
+description = "Update UvPyprojectHandler"
```
srstevenson: Can we make this more explicit? For example, “Support management of package indexes for uv”?

kraken-build/src/kraken/std/python/buildsystem/uv.py
antoinebrl: Here I'm just updating the CLI options as `--index-url` and `--extra-index-url` are marked as deprecated.
The comment is currently still valid because `pyproject-build` can't forward arguments when calling `uv`. As such the configuration is converted into a set of environment variables.

I plan to rework this part in a follow up PR. `uv build` works even when no `build-system` is specified in the `pyproject.toml`.

kraken-build/src/kraken/std/python/buildsystem/poetry_test.py
```diff
@@ -1,4 +1,6 @@
 from kraken.common.toml import TomlFile
+from kraken.std.python.pyproject import PyprojectHandler
+
 from kraken.std.python.buildsystem.poetry import PoetryPyprojectHandler
 from kraken.std.python.pyproject import PackageIndex
```
srstevenson:
```suggestion

from kraken.std.python.buildsystem.poetry import PoetryPyprojectHandler
from kraken.std.python.pyproject import PackageIndex, PyprojectHandler
```
(we should use `ruff check --fix` to automate this: https://docs.astral.sh/ruff/formatter/#sorting-imports)

kraken-build/src/kraken/std/python/buildsystem/uv.py
antoinebrl: I'm updating your suggestion here to represent the optional name with None. However, I decided not to update the `alias` field in the `PackageIndex` dataclass. See my argument bellow: https://github.com/kraken-build/kraken/pull/297#discussion_r1925342139

kraken-build/src/kraken/std/python/buildsystem/poetry_test.py
```diff
@@ -9,14 +11,14 @@
 
 [[tool.poetry.source]]
 name = "foo"
-url = "https://foo.bar/simple/"
+url = "https://foo.com/simple/"
```
srstevenson: If we're changing these, let's take the opportunity to use one of the IANA domains reserved for documentation: https://www.iana.org/help/example-domains

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -62,136 +63,145 @@ def unsafe_url(self) -> str:
...
+                entry["name"] = index.name
+            if index.default:
+                entry["default"] = True
+            if index.explicit:
+                entry["explicit"] = True
+            entries.append(entry)
+        return entries
 
     def to_env(self) -> dict[str, str]:
+        """Convert UV configuration for indexes into a environment variables."""
```
srstevenson:
```suggestion
        """Convert UV configuration for indexes into environment variables."""
```

kraken-build/src/kraken/std/python/buildsystem/uv_test.py
```diff
@@ -0,0 +1,156 @@
+from kraken.common.toml import TomlFile
+from kraken.std.python.pyproject import PyprojectHandler
+
+from kraken.std.python.buildsystem.uv import UvPyprojectHandler, UvIndexes
+from kraken.std.python.pyproject import PackageIndex
```
srstevenson:
```suggestion

from kraken.std.python.buildsystem.uv import UvPyprojectHandler, UvIndexes
from kraken.std.python.pyproject import PackageIndex, PyprojectHandler
```

kraken-build/.changelog/_unreleased.toml
antoinebrl: Done! Updated the PR title too

kraken-build/src/kraken/std/python/buildsystem/poetry_test.py
antoinebrl: That is a bit annoying ...

kraken-build/src/kraken/std/python/buildsystem/poetry_test.py
antoinebrl: I replaced all dummy URLs with `example.com` or `example.org`

kraken-build/src/kraken/std/python/buildsystem/poetry_test.py
srstevenson: I've opened https://github.com/kraken-build/kraken/pull/298/ to automate this.

srstevenson (approved):

PR #299, opened 2025-01-22
Remove `uv` as a direct dependency in kraken-build

antoinebrl:
This mimics how other package managers are operated, that is by assuming they are installed by the user and available via the `PATH` environment variable. This will also reduce potential mismatches between the different installed versions.

Stacked on #297

### DIFF
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -15,3 +15,9 @@ id = "4d8a687f-14ae-4644-904a-80b1aef3421b"
 type = "improvement"
 description = "Support management of package indexes for uv"
 author = "antoine.broyelle@helsing.ai"
+
+[[entries]]
+id = "06add927-62e1-4503-b82f-88000bd29da1"
+type = "improvement"
+description = "Remove `uv` as a direct dependency"
+author = "antoine.broyelle@helsing.ai"
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -19,5 +19,5 @@ author = "antoine.broyelle@helsing.ai"
 [[entries]]
 id = "06add927-62e1-4503-b82f-88000bd29da1"
 type = "improvement"
-description = "Remove `uv` as a direct dependency"
+description = "Remove `uv` as a direct dependency in kraken-build"
 author = "antoine.broyelle@helsing.ai"
--- a/kraken-build/src/kraken/std/python/buildsystem/uv.py
+++ b/kraken-build/src/kraken/std/python/buildsystem/uv.py
@@ -13,10 +13,11 @@ import subprocess as sp
 import tempfile
 from collections.abc import Sequence
 from dataclasses import dataclass
-from os import fsdecode
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Any, Iterable, TypeVar
+from typing import Annotated, Any, Iterable, TypeVar
 
+from kraken.common import NotSet
+from kraken.common.pyenv import get_current_venv
 from kraken.common.sanitize import sanitize_http_basic_auth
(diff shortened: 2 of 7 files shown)


vsiles (approved):
Overall happy with the MR but can we document this behavior to make it easily discoverable ? Something like "kraken is using `uv` from the `$PATH`. If you use a custom uv, make sure it is available in there"

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -294,25 +278,49 @@ def get_lockfile(self) -> Path | None:
...
+        if venv:
+            venv.deactivate(environ)
+
+        command = ["uv", "run", "bash", "echo \$VIRTUAL_ENV"]
+        try:
+            response = sp.check_output(command, cwd=self.project_directory, env=environ).decode().strip()
+        except sp.CalledProcessError as exc:
+            if exc.returncode != 1:
+                raise
+            return None
```
antoinebrl, line 303: I find that the management of the error and returned value between this block and the parent function `get_path` is a bit weird. However, this is inspired by the `PoetryManagedEnvironment` and `PdmManagedEnvironment` so I went this this approach for consistency.

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -294,25 +278,49 @@ def get_lockfile(self) -> Path | None:
...
+        """Uses `uv run` to determines the location of the venv."""
+
+        # Ensure we de-activate any environment that might be active when Kraken is invoked. Otherwise,
+        # Poetry would fall back to that environment.
+        environ = os.environ.copy()
+        venv = get_current_venv(environ)
+        if venv:
+            venv.deactivate(environ)
+
+        command = ["uv", "run", "bash", "echo \$VIRTUAL_ENV"]
```
srstevenson: This might be preferable to avoid the need for escaping, and `printenv` is a binary so we don't need to fork bash:
```suggestion
        command = ["uv", "run", "printenv", "VIRTUAL_ENV"]
```

antoinebrl: This produces an error on my machine:
```
/usr/bin/printenv: /usr/bin/printenv: cannot execute binary file
```

antoinebrl:
> Overall happy with the MR but can we document this behavior to make it easily discoverable ? Something like "kraken is using uv from the $PATH. If you use a custom uv, make sure it is available in there"

@vsiles , I added a note in the documentation based on your suggestion. Let me know what you think.

kraken-build/src/kraken/std/python/buildsystem/uv.py
srstevenson: I can't reproduce on macOS or Ubuntu; but it's not a blocker, and we can retain your original approach.
```
$ uv run printenv USER
scott
```

antoinebrl: Actually, `printenv` works but I forgot the `-c` argument when calling `bash`.

.changelog/_unreleased.toml
```diff
@@ -15,3 +15,9 @@ id = "4d8a687f-14ae-4644-904a-80b1aef3421b"
 type = "improvement"
 description = "Support management of package indexes for uv"
 author = "antoine.broyelle@helsing.ai"
+
+[[entries]]
+id = "06add927-62e1-4503-b82f-88000bd29da1"
+type = "improvement"
+description = "Remove `uv` as a direct dependency"
```
```suggestion
description = "Remove `uv` as a direct dependency in kraken-build"
```

It's kept around in `kraken-wrapper` (and I think that's good!)

srstevenson (approved):

PR #302, opened 2025-01-24
[uv] Inject index environment variables to subprocesses

antoinebrl:
Users may choose to set up their own `uv.toml` global configuration file and define environment variables containing credentials. However, it is Kraken's responsibility to accurately manage the placement and integration of these credentials if the `krakenw auth` command is used. In addition, if there is a need to deploy packages to different repositories, the indexes and their credentials can be dynamically created during runtime. Given that `uv` currently lacks keyring support, I recommend that we inject information about the indexes via environment variables. This prevents potentially leaking the credentials by logging the full command.

I also took this opportunity to prettify and standardize logging messages when the build system managers are running sub-commands.


vsiles:
Nitpicking here, but most of the MR is actually log improvement and not uv related. You might want to split it in two (but I won't fight for it :D)

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -221,7 +220,9 @@ def get_managed_environment(self) -> ManagedEnvironment:
 
     def update_lockfile(self, settings: PythonSettings, pyproject: TomlFile) -> TaskStatus:
         command = ["uv", "lock", "--package"]
-        sp.check_call(command, cwd=self.project_directory)
+        logger.debug("Run '%s' in %s.", " ".join(command), self.project_directory)
+        env_vars = UvIndexes.from_package_indexes(settings.package_indexes.values()).to_env()
```
vsiles: Could we make this a function /method to be sure we never forget to bundle these together ?

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -302,7 +293,7 @@ def _get_uv_environment_path(self) -> Path | None:
                 raise
             return None
 
-        logger.debug("Virtual environment detected: %s", response)
+        logger.info("Virtual environment detected: %s", response)
```
vsiles, line 291: Does this align with the behavior of other build systems ?

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -259,17 +253,14 @@ def build_v2(self, settings: PythonSettings, output_directory: Path) -> list[Pat
...
                 "uv",
             ]
-            logger.info(
-                "Running %s in '%s' with env %s",
-                command,
-                self.project_directory,
-                sanitize_http_basic_auth(str(env)),
-            )
-            sp.check_call(command, cwd=self.project_directory, env={**os.environ, **env})
+            logger.debug("Run '%s' in %s.", " ".join(command), self.project_directory)
```
vsiles: logging level change spotted.  (here and  line 223). In the other files they were `info`. Why the change to `debug`?

vsiles:
Is there anyway to test the behavior ?

kraken-build/src/kraken/std/python/buildsystem/pdm.py
```diff
@@ -132,6 +132,7 @@ def get_managed_environment(self) -> ManagedEnvironment:
 
     def update_lockfile(self, settings: PythonSettings, pyproject: TomlFile) -> TaskStatus:
         command = ["pdm", "update"]
+        logger.info("Run '%s'", " ".join(command))
```
Ghelfi, line 135: Since you are prettifying the logs, can we consider using f-string in the modified files?

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -259,17 +253,14 @@ def build_v2(self, settings: PythonSettings, output_directory: Path) -> list[Pat
...
+
+            env_vars = UvIndexes.from_package_indexes(settings.package_indexes.values()).to_env()
```
Ghelfi: Double tap on @vsiles comment. Would be nice to ship this in a function to mitigate error risk.

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -322,8 +313,9 @@ def get_path(self) -> Path:
 
     def install(self, settings: PythonSettings) -> None:
         command = ["uv", "sync", "--all-extras", "--frozen"]
-        logger.info("%s", command)
-        sp.check_call(command, cwd=self.project_directory)
+        logger.info("Run '%s'", " ".join(command))
+        env_vars = UvIndexes.from_package_indexes(settings.package_indexes.values()).to_env()
```
Ghelfi: ditto

kraken-build/src/kraken/std/python/buildsystem/pdm.py
antoinebrl: The recommended way is to let the logger format the log message himself. The string rendering might be different or skipped depending on the formatter, the log destination, and the log level. It would be possible to lint against f-string in log statements: https://docs.astral.sh/ruff/rules/logging-f-string/

kraken-build/src/kraken/std/python/buildsystem/uv.py
antoinebrl: Great suggestion! I reformatted the code to use an utility function for when index credentials need to be injected.

kraken-build/src/kraken/std/python/buildsystem/uv.py
antoinebrl: Done! see comment above

antoinebrl:
Hi @vsiles ! Thanks for the review. Sorry it took some time to get back to you, I was chasing a nasty bug.

Because the slap changelog produces conflicts on every single merge or rebase I may have tried to squeeze two things in one PR. I will make sure to make atomic PRs in the future.

Regarding testing, there is already a test for that. The `uv-project` is installed, built and published. Then the `uv-project-consumer`, which relies on `uv-project` is installed. The error made was to add `UV_DEFAULT_INDEX` in the test rather than implementing the logic to inject index credentials on every call to `uv`. This is solved now.

Ghelfi (approved):

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -321,9 +307,17 @@ def get_path(self) -> Path:
...
-        command = ["uv", "sync", "--all-extras", "--frozen"]
-        logger.info("%s", command)
-        sp.check_call(command, cwd=self.project_directory)
+        _run_uv_command_with_indexes(["uv", "sync", "--all-extras"], settings, self.project_directory)
 
     def always_install(self) -> bool:
         return True
+
+
+def _run_uv_command_with_indexes(command: list[str], settings: PythonSettings, work_dir: Path) -> None:
```
srstevenson: This doesn't require the command to be a uv subcommand: can we rename it to `_run_with_uv_indexes` or similar?

antoinebrl: Good name suggestion! Implemented

srstevenson (approved):

Issue #300, opened 2025-01-22
Make cli-log-level configurable


rfechtner:
The pytest task seems to hardcode `--log-cli-level INFO`, which will effectively ignore .ini options or `addopts` overrides. 
Could not find an alternative easily limit the terminal logs to e.g. WARNING.

Ref: https://github.com/kraken-build/kraken/blob/5980e4f49722cc030f61cee1ae7aa8ffec12f146/kraken-build/src/kraken/std/python/tasks/pytest_task.py#L72C24-L72C37

PR #304, opened 2025-01-24
Remove --log-cli-level INFO when calling pytest

antoinebrl:
Setting the log level to INFO displays all the test names. If the test suite is large the logs long can be hard to digest. `pytest-xdist` make the problem even worse by printing each test name twice, on start and on completion. Moreover, this argument can conflict with options specified in the `pyproject.toml`, preventing users from adjusting the information displayed by pytest. By removing this option, one would be able to track progress with a progress bar for example while still displaying the test name in the summary table if needed.

Solves https://github.com/kraken-build/kraken/issues/300

### DIFF
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -33,3 +33,9 @@ id = "00ded573-ed92-4aab-b480-e21589718bbd"
 type = "fix"
 description = "Remove used of deprecated flag to be removed in pip 25.1"
 author = "scott@stevenson.io"
+
+[[entries]]
+id = "c34f7131-4e5d-482f-b654-bd08976c59fc"
+type = "fix"
+description = "Remove --log-cli-level INFO when calling pytest. This option can be added back by the users via their configuration files."
+author = "antoine.broyelle@helsing.ai"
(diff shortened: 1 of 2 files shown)


vsiles:
Since this will change the output people are used to, would it be possible to make it explicit in the changelog how to get some of these logs back in the new version ?

srstevenson:
To maintain compatibility for users accustomed to this output format, an alternative approach could be to continue passing the `--log-cli-level` argument while making the level configurable via the pytest task. This would default to `INFO` for compatibility:
```python
pytest_task(cli_log_level="INFO")
```
The tradeoff is that maintaining compatibility requires configuring this differently when using Kraken compared to calling pytest directly.

Tpt:
My 2cents: If it's possible to configure verbosity using `pyproject.toml` I am not sure that having an option in Kraken brings much more than extra complexity. I would not bother too much on backward compatibility, this is only debug verbosity and if the changelog is clear enough (giving an example on how to revert the change), users should be able to migrate smoothly.

antoinebrl:
I'm quite aligned with @Tpt and can only echo what he said. Having that configuration in two places will introduce some complexity, both for the users and some extra logic to fuse the different sources. Add the end, this is only impacting the structure of the logs and the users can easily add back this option via the configuration file. I've updated the description in the changelog.

Tpt (approved):

vsiles (approved):

PR #305, opened 2025-01-25
Use `uv tool` instead of `pex` to run Python tools

srstevenson:
This MR adopts [uv's support for invoking Python tools](https://docs.astral.sh/uv/guides/tools/). This provides fast resolution and caching, and allows us to replace the previous approach of generating pex files for each formatter and linter. The pex build task will be removed in a follow up PR.

### DIFF
--- a/kraken-build/src/kraken/std/python/tasks/black_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/black_task.py
@@ -7,7 +7,6 @@ from pathlib import Path
 from kraken.common import Supplier
 from kraken.core import Project, Property
 from kraken.core.system.task import TaskStatus
-from kraken.std.python.tasks.pex_build_task import pex_build
 
 from .base_task import EnvironmentAwareDispatchTask
 
@@ -17,7 +16,7 @@ class BlackTask(EnvironmentAwareDispatchTask):
 
     python_dependencies = ["black"]
 
-    black_bin: Property[str] = Property.default("black")
+    black_cmd: Property[Sequence[str]] = Property.default(["black"])
     check_only: Property[bool] = Property.default(False)
     config_file: Property[Path]
     additional_args: Property[Sequence[str]] = Property.default_factory(list)
@@ -26,7 +25,7 @@ class BlackTask(EnvironmentAwareDispatchTask):
     # EnvironmentAwareDispatchTask
 
     def get_execute_command_v2(self, env: MutableMapping[str, str]) -> list[str] | TaskStatus:
-        command = [self.black_bin.get(), str(self.settings.source_directory)]
+        command = [*self.black_cmd.get(), str(self.settings.source_directory)]
         command += self.settings.get_tests_directory_as_args()
         command += [str(directory) for directory in self.settings.lint_enforced_directories]
         command += [str(p) for p in self.additional_files.get()]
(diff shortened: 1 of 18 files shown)


antoinebrl:
This is looking very promising! Is your plan to remove `pex` from the list of dependencies and remove the `PexBuildTask` ? (or maybe have optional dependency on `pex` and conditioned imports?)

kraken-build/src/kraken/std/python/tasks/black_task.py
```diff
@@ -64,18 +63,16 @@ def black(
...
     project = project or Project.current()
 
     if version_spec is not None:
-        black_bin = pex_build(
-            "black", requirements=[f"black{version_spec}"], console_script="black", project=project
-        ).output_file.map(str)
+        black_bin = Supplier.of(["uv", "tool", "run", "--from", f"black{version_spec}", "black"])
     else:
-        black_bin = Supplier.of("black")
+        black_bin = Supplier.of(["black"])
```
Ghelfi: I am not a big fan of the current process where we use pex (now uv) to download to linter / formatter if the version is specified and rely on the binary (which need to be locally there) else.

Do we plan to keep this or to use `uv tool run --from ...` even if the version is not specified?

kraken-build/src/kraken/std/python/tasks/black_task.py
```diff
@@ -17,7 +16,7 @@ class BlackTask(EnvironmentAwareDispatchTask):
 
     python_dependencies = ["black"]
 
-    black_bin: Property[str] = Property.default("black")
+    black_bin: Property[list[str]] = Property.default(["black"])
```
Ghelfi: There is an alternance between `list` and `Sequence`. Can we pick one?

Ghelfi (commented):
Looking good!

kraken-build/src/kraken/std/python/tasks/black_task.py
srstevenson: Calling a binary installed outside Kraken when a version isn't specified is a fallback to the behaviour before the pex support was added, where the tool always had to be already installed. We can remove this once uv is used to manage all Python tools, but let's do that atomically in a separate PR.

srstevenson:
> Is your plan to remove `pex` from the list of dependencies and remove the `PexBuildTask`? (or maybe have optional dependency on `pex` and conditioned imports?)

Once we've validated that there are no external consumers of `PexBuildTask` outside `kraken.std`, I see no need to keep the task or the dependency.

kraken-build/src/kraken/std/python/tasks/mypy_task.py
```diff
@@ -49,10 +46,10 @@ def get_execute_command_v2(self, env: MutableMapping[str, str]) -> list[str]:
         status_file = (self.project.directory / ".dmypy.json").absolute()
         if use_daemon:
             command += ["--status-file", str(status_file), "run", "--"]
-        if mypy_pex_bin:
+        if mypy_cmd:
             # Have mypy pick up the Python executable from the virtual environment that is activated automatically
             # during the execution of this task as this is an "EnvironmentAwareDispatchTask". If we don't supply this
-            # option, MyPy will only know the packages in its PEX.
+            # option, MyPy will only know the packages in its own virtual environment.
             command += ["--python-executable", "python"]
```
srstevenson: This needs updating so mypy can access dependencies outside the `uv tool`-managed environment.

kraken-build/src/kraken/std/python/tasks/black_task.py
Ghelfi: Fair enough 👍

Tpt:
Mypy subtest is not migrated. Happy to do it in a follow-up if you prefer to not do it

Tpt (approved):

kraken-build/src/kraken/std/python/tasks/mypy_task.py
```diff
@@ -49,11 +47,11 @@ def get_execute_command_v2(self, env: MutableMapping[str, str]) -> list[str]:
...
-        if mypy_pex_bin:
-            # Have mypy pick up the Python executable from the virtual environment that is activated automatically
-            # during the execution of this task as this is an "EnvironmentAwareDispatchTask". If we don't supply this
-            # option, MyPy will only know the packages in its PEX.
-            command += ["--python-executable", "python"]
+        if mypy_cmd and self.settings.build_system is not None:
+            # Have mypy pick up the Python executable from the managed virtual environment for this project.
+            # If we don't supply this, MyPy will only know the packages in its own virtual environment.
+            managed_env_path = self.settings.build_system.get_managed_environment().get_path()
+            command += ["--python-executable", os.fspath(managed_env_path / "bin" / "python")]
```
antoinebrl, line 54: Another solution might be to specify the python interpreter within the venv dedicated to the tool. Something along this line `uv tool run --python /path/to/venv/bin/python mypy`.

antoinebrl (approved):

kraken-build/src/kraken/std/python/tasks/mypy_task.py
srstevenson: I don't expect that'll work, as the `uv tool run` entrypoint still creates a new virtual environment using that interpreter, which won't have the project dependencies installed in it: https://github.com/astral-sh/uv/blob/ec480bd3ee43da2f5f15e7c6383c334e78894c31/crates/uv/src/commands/tool/run.rs#L118

PR #307, opened 2025-01-27
Directly call `uv build` even when no `[build-system]` is specified in the `pyproject.toml`

antoinebrl:
Extract from [the `uv` documentation on build systems](https://docs.astral.sh/uv/concepts/projects/config/#build-systems): 

> While uv will not build and install the current project without a build system definition, the presence of a [build-system] table is not required in other packages. For legacy reasons, if a build system is not defined, then setuptools.build_meta:__legacy__ is used to build the package. Packages you depend on may not explicitly declare their build system but are still installable. Similarly, if you add a dependency on a local package or install it with uv pip, uv will always attempt to build and install it.

As such, this PR is proposing to remove the reliance on `pyproject-build` in favor of `uv build`.

### DIFF
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -3,3 +3,9 @@ id = "62408706-cb66-49ed-95cf-d44382829ed6"
 type = "hygiene"
 description = "Remove unused dependencies"
 author = "thomas.pellissier-tanon@helsing.ai"
+
+[[entries]]
+id = "18cfdd07-b11d-4d22-906f-bc4080cda316"
+type = "improvement"
+description = "Directly call `uv build` even when no `[build-system]` is specified in the `pyproject.toml`"
+author = "antoine.broyelle@helsing.ai"
--- a/kraken-build/src/kraken/std/python/buildsystem/uv.py
+++ b/kraken-build/src/kraken/std/python/buildsystem/uv.py
@@ -8,9 +8,7 @@ from __future__ import annotations
 
 import logging
 import os
-import shutil
 import subprocess as sp
-import tempfile
 from collections.abc import Sequence
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -230,34 +228,10 @@ def requires_login(self) -> bool:
...
-            _run_uv_command_with_indexes(command, settings, self.project_directory)
-
-            src_files = list(Path(tempdir).iterdir())
-            dst_files = [output_directory / path.name for path in src_files]
-            for src, dst in zip(src_files, dst_files, strict=True):
-                shutil.move(str(src), dst)
-
-        return dst_files
+        _run_uv_command_with_indexes(
+            ["uv", "build", f"--out-dir={output_directory.absolute()}"], settings, self.project_directory
```
antoinebrl, line 284: @Tpt @vsiles `uv build` support multiple backends: `hatch`, `flit`, `pdm`, `setuptools` and `maturin` and `scikit`. Do you think using `uv build` here would remove the need for `MaturinUvPythonBuildSystem`? If that is the case, we would have a unified interface for all things `uv` independently of the build backend.

Tpt: Sadly not: `MaturinUvPythonBuildSystem` takes care of building wheels targetting Python stable ABI (abi3) for multiple platforms using cross-compilation whereas `maturin` called via `uv build` will by default only build a wheel for the current system.

Tpt (commented):
Thanks!

Tpt (approved):

.changelog/_unreleased.toml
```diff
@@ -45,3 +45,9 @@ id = "dddb0086-94f2-4b18-b73d-9816b2fc296f"
 type = "improvement"
 description = "[uv] Inject index environment variables to subprocesses"
 author = "antoine.broyelle@helsing.ai"
+
+[[entries]]
+id = "6a2d1204-cc1d-43f1-9b45-577aec55a104"
+type = "improvement"
+description = "Directly call `uv build` even when no \\[build-system\\] is specified"
```
srstevenson: Was this some escaping to pass as a command line argument to slap?
```suggestion
description = "Directly call `uv build` even when no [build-system] is specified"
```

srstevenson (approved):

kraken-build/src/kraken/std/python/buildsystem/uv.py
vsiles: Agreed. The main goal here is cross-compilation. I don't think uv is providing anything like that out of the box (nor that it should)

kraken-build/src/kraken/std/python/buildsystem/pdm.py
```diff
@@ -132,6 +132,7 @@ def get_managed_environment(self) -> ManagedEnvironment:
 
     def update_lockfile(self, settings: PythonSettings, pyproject: TomlFile) -> TaskStatus:
         command = ["pdm", "update"]
+        logger.info("Run '%s'", " ".join(command))
         sp.check_call(command, cwd=self.project_directory)
```
vsiles: Nit: we would really benefit to provide a single function for all these calls ;)

vsiles (approved):

kraken-build/src/kraken/std/python/buildsystem/uv.py
antoinebrl: Thanks for the clarification! `uv build` has the `--config-setting` which are passed down to the build backend if that helps.

PR #309, opened 2025-01-29
Avoid reading pyproject.toml when installing Kraken venv to avoid conflicts between indexes

Tpt:
By default `uv pip` reads the local `pyproject.toml` and uses indexes declared there. Kraken has its own indexes configuration format and we want to avoid conflicts between them

### DIFF
--- a/kraken-wrapper/src/kraken/wrapper/_buildenv_uv.py
+++ b/kraken-wrapper/src/kraken/wrapper/_buildenv_uv.py
@@ -32,7 +32,7 @@ class UvBuildEnv(_VenvBuildEnv):
 
     def _get_install_command(self, venv_dir: Path, requirements: RequirementSpec, env: dict[str, str]) -> list[str]:
         env["VIRTUAL_ENV"] = str(venv_dir)
-        return [self._uv_bin, "pip", "install", *requirements.to_args(base_dir=self._project_root)]
+        return [self._uv_bin, "pip", "install", "--no-config", *requirements.to_args(base_dir=self._project_root)]
 
     def get_type(self) -> EnvironmentType:
         return EnvironmentType.UV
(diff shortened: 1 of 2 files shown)


kraken-wrapper/src/kraken/wrapper/_buildenv_uv.py
```diff
@@ -32,7 +32,7 @@ def _get_create_venv_command(self, python_bin: Path, path: Path) -> list[str]:
 
     def _get_install_command(self, venv_dir: Path, requirements: RequirementSpec, env: dict[str, str]) -> list[str]:
         env["VIRTUAL_ENV"] = str(venv_dir)
-        return [self._uv_bin, "pip", "install", *requirements.to_args(base_dir=self._project_root)]
+        return [self._uv_bin, "pip", "install", "--no-config", *requirements.to_args(base_dir=self._project_root)]
```
Ghelfi, line 35: Would it be possible to test this?

Ghelfi (approved):
LGTM!

kraken-wrapper/src/kraken/wrapper/_buildenv_uv.py
Tpt: UV only warns when it can't access an index so we would need to build a corner case to properly fail if the option is not set. I am not sure it's worth the time (I got an error locally because of this issue but I am not sure what is actually the cause of it so it's hard for me to create a minimal test to reproduce it).

Ghelfi: Would something like this help? (from [warnings doc](https://docs.python.org/3/library/warnings.html#warnings.filterwarnings))
```
warnings.filterwarnings("error", category=CorrespondingWarningType)
```

Not saying we should do it now, just for my curiosity.

Tpt: The warning is printed by UV that is in a child process, not sure it will help

antoinebrl:
Good catch!

antoinebrl (approved):

PR #319, opened 2025-02-04
Removes cargo_sqlx and cargo_deb tasks

Tpt:
Unused internally and on GitHub


vsiles (approved):
Not against cleanup, but can we give a motivation for that move in the description ? (like "we no longer/never used it" ?

Tpt:
> Not against cleanup, but can we give a motivation for that move in the description ? (like "we no longer/never used it" ?

Done! I don't think it got actually used anywhere and code search does not yield anything now

antoinebrl (approved):

srstevenson (approved):

PR #322, opened 2025-02-05
pytest: don't scan `src/`

antoinebrl:
Kraken defaults orient the users towards having source files under `src/` and tests under `tests/`. Pytest don't need to know where the source code is, it should be installed (as editable) in the virtual environment in which the tests are executed.

I stumbled upon this problem when working with an optional dependency. The tests were skipped as expected but `krakenw run :python.test.unit` was failing. The submodule built around the optional dependency was scanned by `pytest` and imports raised errors.

### DIFF
--- a/kraken-build/src/kraken/std/python/tasks/pytest_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/pytest_task.py
@@ -61,10 +61,20 @@ class PytestTask(EnvironmentAwareDispatchTask):
                 DeprecationWarning,
             )
 
+        if (
+            self.settings.source_directory not in tests_dir
+            and self.settings.source_directory not in self.include_dirs.get()
+            and list((self.project.directory / self.settings.source_directory).rglob("*_test.py"))
+        ):
+            warnings.warn(
+                f"Some test files are detected inside the source directory '{self.settings.source_directory }'. "
+                f"These tests will be skipped. Consider adding '{self.settings.source_directory }' to `tests_dir`.",
+                DeprecationWarning,
+            )
+
         command = [
             "pytest",
             "-vv",
-            str(self.project.directory / self.settings.source_directory),
             *[str(self.project.directory / path) for path in tests_dir],
             *[str(self.project.directory / path) for path in self.include_dirs.get()],
         ]
--- a/kraken-build/src/kraken/std/python/tasks/pytest_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/pytest_task.py
@@ -67,8 +67,8 @@ class PytestTask(EnvironmentAwareDispatchTask):
             and list((self.project.directory / self.settings.source_directory).rglob("*_test.py"))
         ):
             warnings.warn(
-                f"Some test files are detected inside the source directory '{self.settings.source_directory }'. "
-                f"These tests will be skipped. Consider adding '{self.settings.source_directory }' to `tests_dir`.",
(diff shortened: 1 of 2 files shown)


NiklasRosenstein:
My personal preference is to keep tests for a specific module next to it (e.g. `mymodule_test.py`), and only put broader tests into the `tests/` directory. Making it more clever to not scan the `src/` directory in some case or making it configurable would be appreciated.

antoinebrl:
> Making it more clever to not scan the `src/` directory in some case or making it configurable would be appreciated.

Hi! The `PytestTask` already takes multiple options regarding the folders to scan: `tests_dir: Sequence[Path] | Path` and `include_dirs: Sequence[Path]`. As such, the users can already configure it if they want to place test files under `src/`. Do you have something else in mind ?

NiklasRosenstein:
You're right, that's good enough. Btw. one of those two options should probably be removed at some point 🤔 

Did you think about how to roll this out safely without projects suddenly missing their tests in `src/`?

antoinebrl:
Using Github search, there are no public projects making use of Kraken. Internally, only 3 repos have test files inside `src/`. I added a warning with instructions on how to configure `pytest` in this case.

Ghelfi (approved):

kraken-build/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -61,10 +61,20 @@ def get_execute_command_v2(self, env: MutableMapping[str, str]) -> list[str] | T
...
             )
 
+        if (
+            self.settings.source_directory not in tests_dir
+            and self.settings.source_directory not in self.include_dirs.get()
+            and list((self.project.directory / self.settings.source_directory).rglob("*_test.py"))
+        ):
+            warnings.warn(
+                f"Some test files are detected inside the source directory '{self.settings.source_directory }'. "
+                f"These tests will be skipped. Consider adding '{self.settings.source_directory }' to `tests_dir`.",
```
srstevenson:
```suggestion
                f"Some test files are detected inside the source directory '{self.settings.source_directory}'. "
                f"These tests will be skipped. Consider adding '{self.settings.source_directory}' to `tests_dir`.",
```

kraken-build/src/kraken/std/python/tasks/pytest_task.py
```diff
@@ -61,10 +61,20 @@ def get_execute_command_v2(self, env: MutableMapping[str, str]) -> list[str] | T
                 DeprecationWarning,
             )
 
+        if (
+            self.settings.source_directory not in tests_dir
+            and self.settings.source_directory not in self.include_dirs.get()
+            and list((self.project.directory / self.settings.source_directory).rglob("*_test.py"))
```
srstevenson, line 67: Do you need to search for `test_*.py` too?

antoinebrl: When tests are located under `src/` the goal is to have them nearby the relevant code. Therefore, the `test` is added as a suffix and not a prefix to preserve the alphabetical ordering.

srstevenson: Conventionally, yes, but pytest's test discovery supports either. I don't mind us not supporting this, so long as it's a conscious choice based on looking at Kraken-using codebases rather than an oversight.

antoinebrl: This is on purpose!

Issue #293, opened 2025-01-03
`cargo_publish` tasks fail with `cargo 1.83.0`


ulucs:
cargo 1.83.0 has introduced a new behavior that colludes with how kraken publishes rust packages: up until 1.82, changing the version in`cargo.toml` did not result in changes in `Cargo.lock`, but now it does; consequently causing the `cargo publish` command called by kraken to fail with errors regarding changes to lockfiles.

One mitigation I have found is to set `cargo_publish_task.locked = False` for cargo projects with single packages, but this approach also fails if there is more than one package that will be published (ie workspaces).

Tpt:
Fixed by #328

PR #328, opened 2025-02-25
Cargo publish: properly convert path dependencies

Tpt:
Use only the already published package instead of a hybrid mix of the local version and the published package

Issue #293


kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
```diff
@@ -169,24 +210,23 @@ def publish_lib_and_build_app(repository: CargoRepositoryWithAuth, tempdir: Path
...
+            detach=True,
+        )
 
-    host = "localhost"
-    port = container.ports["35504/tcp"][0]["HostPort"]
-    index_url = f"http://{host}:{port}/git"
-    logger.info("Started local cargo registry at %s", index_url)
-    time.sleep(5)
-    return index_url
+        index_url = "http://0.0.0.0:35504/git"
```
Tpt, line 233: No need to introspect the host and port, we have hardcoded them

kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
```diff
@@ -169,24 +215,26 @@ def publish_lib_and_build_app(repository: CargoRepositoryWithAuth, tempdir: Path
...
-        ],
-        ports=["35504:35504"],
-        detach=True,
-    )
+def private_registry(docker_service_manager: DockerServiceManager) -> Iterator[str]:
+    with file_lock(
+        Path("/tmp/kraken_cargo_private_registry_lock")
+    ):  # We hardcode a port, only a single instance must exist at the same time
+        container = docker_service_manager.run(
+            "ghcr.io/d-e-s-o/cargo-http-registry:sha-2edffd8",  # TODO(Tpt): hardcoded because latest docker images are broken
```
Tpt, line 223: Example of CI from develop that fails because of that: https://github.com/kraken-build/kraken/actions/runs/13418579166/job/37803567596

kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
```diff
@@ -169,24 +215,26 @@ def publish_lib_and_build_app(repository: CargoRepositoryWithAuth, tempdir: Path
...
-    host = "localhost"
-    port = container.ports["35504/tcp"][0]["HostPort"]
-    index_url = f"http://{host}:{port}/git"
-    logger.info("Started local cargo registry at %s", index_url)
-    time.sleep(5)
-    return index_url
+        index_url = "http://0.0.0.0:35504/git"
+        logger.info("Started local cargo registry at %s", index_url)
+        time.sleep(5)
+        yield index_url
```
vsiles, line 236: I'm definitely not a python/async expert but it's weird to see a `yield` followed by some code in a function that doesn't loop. Am I missing something ?

vsiles: But then it's a fixture some I'm out of my depth :p

vsiles (approved):
Lots of test for a single line fix. I love it ;)

kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
Tpt: It's a fun way to allow running some destruction code after the test itself has executed. The calling code from pytest does a `next` call to get the fixture output then an other `next` call to run the "destructor"

vsiles: ok ! I was missing the second call. It don't know if I like or is afraid by that :p

PR #329, opened 2025-02-25
CI: Uses fixed cargo-http-registry version

Tpt:
The latest cargo-http-registry image is broken, breaking CI

### DIFF
--- a/kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
+++ b/kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
@@ -171,7 +171,7 @@ def publish_lib_and_build_app(repository: CargoRepositoryWithAuth, tempdir: Path
 @pytest.fixture(scope="session")
 def private_registry(docker_service_manager: DockerServiceManager) -> str:
     container = docker_service_manager.run(
-        "ghcr.io/d-e-s-o/cargo-http-registry:latest",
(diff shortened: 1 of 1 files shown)


kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
```diff
@@ -171,13 +171,13 @@ def publish_lib_and_build_app(repository: CargoRepositoryWithAuth, tempdir: Path
...
     container = docker_service_manager.run(
-        "ghcr.io/d-e-s-o/cargo-http-registry:latest",
+        "ghcr.io/d-e-s-o/cargo-http-registry:sha-2edffd8",  # TODO(Tpt): hardcoded because latest docker images are broken
         [
             "/tmp/test-registry",
             "--addr",
             "0.0.0.0:35504",
         ],
-        ports=["35504:35504"],
+        ports=["35504"],
```
Tpt, line 180: We load the assigned port just below, there is no reason to hardcode the port imho

kraken-build/tests/kraken_std/integration/cargo/test_cargo_private_registry.py
```diff
@@ -181,9 +181,7 @@ def private_registry(docker_service_manager: DockerServiceManager) -> str:
         detach=True,
     )
 
-    host = "localhost"
-    port = container.ports["35504/tcp"][0]["HostPort"]
-    index_url = f"http://{host}:{port}/git"
+    index_url = f"http://0.0.0.0:35504/git"
```
Tpt, line 184: We hardcode the port (see line 180) and the host (line 178), there is no point to do introspection

PR #331, opened 2025-03-05
Disable locking and all extras when using uv for python.install

JAEarly:
The functionality for python.install is now different when using uv rather than poetry.

With poetry, python.install would not change the lock file(s). However, with uv, every command updates the lockfile, changing the functionality of python.install.

IMO the original functionality was better, with a clear separation between python.update and python.install.

### DIFF
--- a/kraken-build/src/kraken/std/python/buildsystem/uv.py
+++ b/kraken-build/src/kraken/std/python/buildsystem/uv.py
@@ -337,7 +337,12 @@ class UvManagedEnvironment(ManagedEnvironment):
         return self._env_path
 
     def install(self, settings: PythonSettings) -> None:
-        _run_with_uv_indexes(["uv", "sync", "--locked"], settings, self.project_directory)
+        # --locked is used only when uv.lock is present
+        _run_with_uv_indexes(
+            ["uv", "sync"] + (["--locked"] if (self.project_directory / "uv.lock").is_file() else []),
+            settings,
+            self.project_directory,
+        )
 
     def always_install(self) -> bool:
         return True
--- a/kraken-build/src/kraken/std/python/buildsystem/uv.py
+++ b/kraken-build/src/kraken/std/python/buildsystem/uv.py
@@ -338,11 +338,12 @@ class UvManagedEnvironment(ManagedEnvironment):
 
     def install(self, settings: PythonSettings) -> None:
         # --locked is used only when uv.lock is present
-        cmd = ["uv", "sync"] 
-        if (self.project_directory / "uv.lock").is_file()
+        cmd = ["uv", "sync"]
+        if (self.project_directory / "uv.lock").is_file():
             cmd.append("--locked")
+
         _run_with_uv_indexes(
-           cmd,
+            cmd,
             settings,
             self.project_directory,
         )
--- a/kraken-build/src/kraken/std/python/buildsystem/uv.py
+++ b/kraken-build/src/kraken/std/python/buildsystem/uv.py
@@ -337,7 +337,7 @@ class UvManagedEnvironment(ManagedEnvironment):
         return self._env_path
 
     def install(self, settings: PythonSettings) -> None:
-        _run_with_uv_indexes(["uv", "sync", "--all-extras", "--no-sync"], settings, self.project_directory)
+        _run_with_uv_indexes(["uv", "sync", "--all-extras", "--locked"], settings, self.project_directory)
 
     def always_install(self) -> bool:
         return True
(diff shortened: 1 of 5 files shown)


kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -337,7 +337,7 @@ def get_path(self) -> Path:
         return self._env_path
 
     def install(self, settings: PythonSettings) -> None:
-        _run_with_uv_indexes(["uv", "sync", "--all-extras"], settings, self.project_directory)
+        _run_with_uv_indexes(["uv", "sync", "--all-extras", "--no-sync"], settings, self.project_directory)
```
JAEarly:
```suggestion
        _run_with_uv_indexes(["uv", "sync", "--all-extras", "--locked"], settings, self.project_directory)
```

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -337,7 +337,7 @@ def get_path(self) -> Path:
         return self._env_path
 
     def install(self, settings: PythonSettings) -> None:
-        _run_with_uv_indexes(["uv", "sync", "--all-extras"], settings, self.project_directory)
+        _run_with_uv_indexes(["uv", "sync", "--all-extras", "--locked"], settings, self.project_directory)
```
antoinebrl: Should we remove the `--all-extras` in favor of the `default-groups`? the extra are removed on the first `uv run` anyway as the venv is sync.
```suggestion
        _run_with_uv_indexes(["uv", "sync", "--locked"], settings, self.project_directory)
```

JAEarly: I'm happy with this, done!

srstevenson:
@JAEarly Can we add a changelog entry for this, please? You can do so with slap: https://niklasrosenstein.github.io/slap/commands/changelog/#slap-changelog-add

antoinebrl (approved):
Thanks!

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -337,7 +337,7 @@ def get_path(self) -> Path:
         return self._env_path
 
     def install(self, settings: PythonSettings) -> None:
-        _run_with_uv_indexes(["uv", "sync", "--all-extras"], settings, self.project_directory)
+        _run_with_uv_indexes(["uv", "sync", "--locked"], settings, self.project_directory)
```
Ghelfi: Why using `--locked` and not `--frozen`?

Ghelfi (approved):

kraken-build/src/kraken/std/python/buildsystem/uv.py
Ghelfi: Actually `--locked` makes more sense :+1:

Ghelfi: The locked option requires the lock to exist.
In the test suite the lock file cannot be generated for the `uv-project-consumer` and `rust-uv-project-consumer` because they depend on the non-consumer projects. Tests cannot pass.

Shall we use the `self.project_directory` there to add the `--locked` only when a `uv.lock` exists?

antoinebrl: That's an excellent observation! Although including the lock file in the repository is a best practice, it's not always present. I'm confident that `uv` can handle the absence of the lock file, even when `--locked` is used. For our purposes, as you suggested, we can conditionally apply this flag based on the presence of `uv.lock`.

Alternatively, we could generate the lock file during testing when the sub-packages are accessible in the local index, but this might significantly disrupt the primary objective of testing the successful installation path.

Ghelfi: I'd be in favor of applying the `--locked` only when a the lock is present. As you said, it is not mandatory per se even if best practice.

kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -337,7 +337,12 @@ def get_path(self) -> Path:
         return self._env_path
 
     def install(self, settings: PythonSettings) -> None:
-        _run_with_uv_indexes(["uv", "sync", "--all-extras"], settings, self.project_directory)
+        # --locked is used only when uv.lock is present
+        _run_with_uv_indexes(
+            ["uv", "sync"] + (["--locked"] if (self.project_directory / "uv.lock").is_file() else []),
```
antoinebrl: personnal preference:
```suggestion
        # --locked is used only when uv.lock is present
        cmd = ["uv", "sync"] 
        if (self.project_directory / "uv.lock").is_file()
            cmd.append("--locked")
        _run_with_uv_indexes(
           cmd,
```
reformatting might be needed

PR #333, opened 2025-03-11
Deprecate python.(black|flake8|isort|pycln|pylint|pypgrade)

Tpt:


kraken-build/src/kraken/std/python/tasks/black_task.py
```diff
@@ -24,6 +25,13 @@ class BlackTask(EnvironmentAwareDispatchTask):
 
     # EnvironmentAwareDispatchTask
 
+    def __init__(self, name: str, project: Project) -> None:
+        super().__init__(name, project)
+        warnings.warn(
+            "python.black will be removed in a future version. Please use python.ruff instead.",
+            DeprecationWarning,
```
srstevenson, line 32: Should we pass a `stacklevel` argument, so that the warning points to the callsite of the `black` function rather than to `BlackTask.__init__`?

Tpt: Done, `stacklevel=4` because `black` calls `project.task` that itself calls `BlackTask.__init__` that itself calls `warnings.warn`

srstevenson (approved):

antoinebrl (approved):
Thanks! it's a great step forward

PR #342, opened 2025-06-06
Document that UV is now the default installer type

srstevenson:
This is a follow up to https://github.com/kraken-build/kraken/pull/281/ that documents that UV is now the default installer type, superseding the VENV installer.


antoinebrl (approved):

Tpt (approved):

Tpt:
I would tend to think we can deprecate non-UV installation method

NiklasRosenstein (approved):

NiklasRosenstein:
> I would tend to think we can deprecate non-UV installation method

https://github.com/kraken-build/kraken/pull/346

vsiles (approved):
Yes please !

NiklasRosenstein:
I'll remove this part of the doc entirely in https://github.com/kraken-build/kraken/pull/349 since it's not longer going to be an option.

PR #343, opened 2025-06-25
Migrate `pyproject.toml` to UV and use a UV workspace

NiklasRosenstein:
Drop Poetry as the build system for the `kraken-build` and `kraken-wrapper` packages and use Hatchling + Uv instead.


kraken-build/src/kraken/std/dist.py
```diff
@@ -175,7 +175,7 @@ def __exit__(self, *a: Any) -> None:
 
 class TarArchiveWriter(ArchiveWriter):
     def __init__(self, path: Path, type_: Literal["", "gz", "bz2", "xz"]) -> None:
-        self._archive = tarfile.open(path, mode="w:" + type_)
+        self._archive = tarfile.open(path, mode="w:" + type_)  # type: ignore[arg-type]
```
NiklasRosenstein:
```suggestion
        self._archive = tarfile.open(path, mode="w:" + type_)  # type: ignore[call-overload]
```

The suggestion by Copilot was incorrect.

.python-version
```diff
@@ -0,0 +1 @@
+3.13.5
```
Tpt, line 1: Nit: do we need it? I would tend to think just relying on `pyproject.toml` `requires-python` is less error-prone.

kraken-build/pyproject.toml
```diff
@@ -1,75 +1,59 @@
 [build-system]
-build-backend = "poetry.core.masonry.api"
-requires = ["poetry-core"]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
Tpt, line 3: Beware, hatchling does not always include not *.py type. It might be worth to check that `gitignore-io-tokens.json.gz` is still included in the built wheels.

Aside tip: if you want to be cutting-edge UV build backend is now stable: https://docs.astral.sh/uv/concepts/build-backend/

kraken-build/pyproject.toml
```diff
@@ -1,75 +1,59 @@
...
 version = "0.44.1"
+description = "Kraken is an extensible build orchestration system."
+readme = "README.md"
+requires-python = ">=3.10,<3.14"
+license = "MIT"
+authors = [
+  {name = "Niklas Rosenstein", email = "rosensteinniklas@gmail.com"},
+]
+dependencies = [
+  "databind>=4.5.0,<5.0.0",
```
Tpt, line 19: tip (please ignore): 
```suggestion
  "databind~=4.5",
  "dill~=0.3.8",
```

Tpt (approved):

.python-version
NiklasRosenstein: Hm, not strictly, what it does is make sure everyone's on the same Python version when developing with UV. Whether that's desirable or not, I have arguments either way. 😆

kraken-build/pyproject.toml
NiklasRosenstein: Good flag. Checked, it's there :)

```console
$ uv build
$ tar -tf dist/kraken_build-0.44.1.tar.gz  | grep gz
kraken_build-0.44.1/src/kraken/std/git/gitignore/data/gitignore-io-tokens.json.gz
```

PR #352, opened 2025-07-09
Improve typing of `Supplier.get_or()` and `Property.__set__()`, update docs of `Supplier.of()` to cover some special cases and (how to) deal with them

NiklasRosenstein:


kraken-build/src/kraken/common/supplier.py
```diff
@@ -37,7 +37,13 @@ def derived_from(self) -> Iterable["Supplier[Any]"]:
...
         """Return the value of the supplier. Depending on the implemenmtation, this may defer to other suppliers."""
 
-    def get_or(self, fallback: U) -> "T | U":
+    @overload
+    def get_or(self, fallback: None) -> "T | None": ...
+
+    @overload
+    def get_or(self, fallback: U) -> "T | U": ...
+
+    def get_or(self, fallback: U | None) -> "T | U | None":
```
NiklasRosenstein, line 46: Before this, the following would fail:

```py
p: Property[int | None]
p.get_or(None) # Argument 1 to "get_or" of "Supplier" has incompatible type "None"; expected "int" [arg-ty[e]
```

PR #374, opened 2025-07-21
fix: Pass `--python-version` argument to `uv pip install` to ensure the minimum interpreter constraint in the buildscript requirements is respected

NiklasRosenstein:
Before this PR, `uv pip install` would be run without the `--python-version` option, making it install the latest package versions compatible with the current Python version, which may not always be the minimum Python version that is permitted by the build script.

The `buildscript(interpreter_constraint)` defaults to `>=3.10`, but when the environment on your machine is getting created with 3.15 for example, you may get `networkx 3.5` which does not install into 3.10.

### DIFF
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -3,3 +3,9 @@ id = "aa2cba40-ff5c-42b0-a1b8-636f522db101"
 type = "fix"
 description = "`krakenw --upgrade` now passes the `--upgrade` flag to `uv pip install`, as expected"
 author = "@NiklasRosenstein"
+
+[[entries]]
+id = "e4028844-450c-4c4c-b01c-cabc0b1ffcf8"
+type = "fix"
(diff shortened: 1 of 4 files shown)


NiklasRosenstein:
This causes issues with native dependencies.

Despite the UV help saying the following:

```
      --python-version <PYTHON_VERSION>        The minimum Python version that should be supported by the requirements (e.g., `3.7` or `3.7.9`)
```

Which seems to suggest that it will pick to install _versions_ of packages that are minimum compatible with the specified Python version, it actually also selects the specific distribution for the specified Python version.

In the case I'm running into, it installs `rpds-py` for Python 3.10 despite the runtime version of the build environment being 3.13, because we pass `--python-version 3.10`.

I think the solution here is to switch to a proper `uv.lock` file for the Kraken build environment.

NiklasRosenstein:
^ Addressed in #376

PR #382, opened 2025-08-07
Ignore error code 403 in additional indexes

ulucs:
Whenever we use multiple indexes with authentication, `uv` might try to search non-existent packages in the additional indexes.

This leads to error codes like
```
➜ curl https://private-index.com/simple/hatchling -L
{
  "errors" : [ {
    "status" : 403,
    "message" : "{\"error\":\"Forbidden\",\"reason\":\"User has no permissions to resolve packages.\"}"
  } ]
}%  
```

This will completely prevent packages from building if a deterministic build-system selection is chosen; so some users opt to using nondeterministic methods like `index-strategy = "unsafe-first-match"`.

Ignoring 403's on extra indexes will allow us to fully search all the indexes as per uv's documentation: https://docs.astral.sh/uv/concepts/indexes/#ignoring-error-codes-when-searching-across-indexes

### DIFF
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -1,0 +1,5 @@
+[[entries]]
+id = "10af1349-1542-4eb8-9680-f25863c57a25"
+type = "fix"
+description = "Ignore error code 403 in additional indexes"
+author = "uluc.sengil@helsing.ai"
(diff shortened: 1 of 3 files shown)


srstevenson (approved):

NiklasRosenstein:
@danilobuerger FYI, this one will also cause our Renovate jobs that upgrade Kraken to fail since we don't run `krakenw run apply` anymore.

PR #389, opened 2025-09-23
Migrate from twine upload to uv publish

John-P:
The `python.publish` task has been updated to use `uv publish` instead of `twine upload`.

Key changes:

- The `PublishTask` now constructs and executes a `uv publish` command.
- Authentication is handled via `UV_USERNAME` and `UV_PASSWORD` environment variables, which is `uv`'s standard practice.
- ~~A new `check_url` setting has been added to `PackageIndex` to support `uv`'s ability to check for existing packages against a different URL than the publishing repository.~~ This is now set in the call to uv using the index url (which was previously unused by the python publish task) when `skip_existing` is true.
- The `interactive` property has been ~~removed~~ deprecated.
  - `uv` has no interactive option, but will be interactive if not credentials are given.
- Tests and documentation have been updated to reflect the switch to `uv`.

### DIFF
--- a/kraken-build/src/kraken/std/python/settings.py
+++ b/kraken-build/src/kraken/std/python/settings.py
@@ -25,6 +25,9 @@ class PythonSettings:
         #: An alternative URL to upload packages to.
         upload_url: str | None
 
+        #: An alternative URL to index_url to check for existing files to skip duplicate uploads
+        check_url: str | None
+
         #: Credentials to use when publishing to or reading from the index.
         credentials: tuple[str, str] | None
 
@@ -78,6 +81,7 @@ class PythonSettings:
         *,
         index_url: str | None = None,
         upload_url: str | None = None,
+        check_url: str | None = None,
         credentials: tuple[str, str] | None = None,
         is_package_source: bool = True,
         priority: PackageIndex.Priority | str = PackageIndex.Priority.supplemental,
@@ -90,6 +94,8 @@ class PythonSettings:
             If not specified, *alias* must be a known package index (`pypi` or `testpypi`).
         :param upload_url: If the upload url deviates from the registry URL. Otherwise, the upload URL will
             be the same as the *url*,
+        :param check_url: Optional index URL to check for existing files. Otherwise, the index URL will be
+            used.
         :param credentials: Optional credentials to read from the index.
         :param is_package_source: If set to `False`, the index will not be used to source packages from, but
             can be used to publish to.
@@ -126,6 +132,7 @@ class PythonSettings:
         self.package_indexes[alias] = self._PackageIndex(
             alias=alias,
             index_url=index_url,
+            check_url=check_url,
             priority=priority,
             upload_url=upload_url,
             credentials=credentials,
--- a/kraken-build/src/kraken/std/python/settings.py
+++ b/kraken-build/src/kraken/std/python/settings.py
@@ -25,9 +25,6 @@ class PythonSettings:
         #: An alternative URL to upload packages to.
         upload_url: str | None
 
-        #: An alternative URL to index_url to check for existing files to skip duplicate uploads
-        check_url: str | None
-
         #: Credentials to use when publishing to or reading from the index.
         credentials: tuple[str, str] | None
 
@@ -81,7 +78,6 @@ class PythonSettings:
         *,
         index_url: str | None = None,
         upload_url: str | None = None,
-        check_url: str | None = None,
         credentials: tuple[str, str] | None = None,
         is_package_source: bool = True,
         priority: PackageIndex.Priority | str = PackageIndex.Priority.supplemental,
@@ -94,8 +90,6 @@ class PythonSettings:
             If not specified, *alias* must be a known package index (`pypi` or `testpypi`).
         :param upload_url: If the upload url deviates from the registry URL. Otherwise, the upload URL will
             be the same as the *url*,
-        :param check_url: Optional index URL to check for existing files. Otherwise, the index URL will be
-            used.
         :param credentials: Optional credentials to read from the index.
         :param is_package_source: If set to `False`, the index will not be used to source packages from, but
             can be used to publish to.
@@ -132,7 +126,6 @@ class PythonSettings:
         self.package_indexes[alias] = self._PackageIndex(
             alias=alias,
             index_url=index_url,
-            check_url=check_url,
             priority=priority,
             upload_url=upload_url,
             credentials=credentials,
--- a/kraken-build/src/kraken/std/python/tasks/publish_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/publish_task.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
+import os
 import subprocess
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable
 from pathlib import Path
 
-from kraken.common import Supplier
 from kraken.core import Project, Property, Task, TaskRelationship
 from kraken.core.system.task import TaskStatus
 
@@ -12,15 +12,15 @@ from ..settings import python_settings
 
 
 class PublishTask(Task):
-    """Publishes Python distributions to one or more indexes using :mod:`twine`."""
+    """Publishes Python distributions to one or more indexes using `uv publish`."""
 
     description = "Upload the distributions of your Python project. [index url: %(index_upload_url)s]"
-    twine_cmd: Property[Sequence[str]]
     index_upload_url: Property[str]
+    index_index_url: Property[str]
+    index_check_url: Property[str | None] = Property.default(None)
     index_credentials: Property[tuple[str, str] | None] = Property.default(None)
     distributions: Property[list[Path]]
     skip_existing: Property[bool] = Property.default(False)
-    interactive: Property[bool] = Property.default(True)
     dependencies: list[Task]
 
     def __init__(self, name: str, project: Project) -> None:
@@ -33,31 +33,28 @@ class PublishTask(Task):
 
     def execute(self) -> TaskStatus:
         credentials = self.index_credentials.get()
-        repository_url = self.index_upload_url.get().rstrip("/") + "/"
         command = [
-            *self.twine_cmd.get(),
-            "upload",
-            "--repository-url",
-            repository_url,
-            "--verbose",
+            "uv",
+            "publish",
+            "--publish-url",
+            self.index_upload_url.get(),
             *[str(x.absolute()) for x in self.distributions.get()],
         ]
-        if credentials:
-            command += [
-                "--username",
-                credentials[0],
-                "--password",
-                credentials[1],
-            ]
-        if not self.interactive.get():
-            command.append("--non-interactive")
         if self.skip_existing.get():
-            command.append("--skip-existing")
+            command.extend(["--check-url", self.index_check_url.get() or self.index_index_url.get()])
 
-        safe_command = [x.replace(credentials[1], "MASKED") for x in command] if credentials else command
+        env = os.environ.copy()
+        if credentials:
+            # See https://docs.astral.sh/uv/guides/package/#publishing-your-package
+            # NOTE: PyPI does not support publishing with username and password anymore,
+            #       should we log a warning if the username is not __token__?
+            env["UV_PUBLISH_USERNAME"] = credentials[0]
+            env["UV_PUBLISH_PASSWORD"] = credentials[1]
+
+        safe_command = command
         self.logger.info("$ %s", safe_command)
 
-        returncode = subprocess.call(command, cwd=self.project.directory)
+        returncode = subprocess.call(command, cwd=self.project.directory, env=env)
         return TaskStatus.from_exit_code(safe_command, returncode)
 
 
@@ -66,12 +63,10 @@ def publish(
     package_index: str,
     distributions: list[Path] | Property[list[Path]],
     skip_existing: bool = False,
-    interactive: bool = True,
     name: str = "python.publish",
     group: str | None = "publish",
     project: Project | None = None,
     after: list[Task] | None = None,
-    twine_version: str = ">=6.0.0,<7.0.0",
 ) -> PublishTask:
     """Create a publish task for the specified registry."""
 
@@ -80,15 +75,13 @@ def publish(
     if package_index not in settings.package_indexes:
         raise ValueError(f"package index {package_index!r} is not defined")
 
-    twine_cmd = Supplier.of(["uv", "tool", "run", "--from", f"twine{twine_version}", "twine"])
-
     index = settings.package_indexes[package_index]
     task = project.task(name, PublishTask, group=group)
-    task.twine_cmd = twine_cmd
     task.index_upload_url = index.upload_url
+    task.index_index_url = index.index_url
+    task.index_check_url = index.check_url
     task.index_credentials = index.credentials
     task.distributions = distributions
     task.skip_existing = skip_existing
-    task.interactive = interactive
     task.depends_on(*(after or []))
     return task
--- a/kraken-build/src/kraken/std/python/tasks/publish_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/publish_task.py
@@ -77,6 +77,7 @@ def publish(
     package_index: str,
     distributions: list[Path] | Property[list[Path]],
     skip_existing: bool = False,
+    interactive: bool | None = None,
     name: str = "python.publish",
     group: str | None = "publish",
     project: Project | None = None,
@@ -97,5 +98,6 @@ def publish(
     task.index_credentials = index.credentials
     task.distributions = distributions
     task.skip_existing = skip_existing
+    task.interactive = interactive
     task.depends_on(*(after or []))
     return task
--- a/kraken-build/src/kraken/std/python/tasks/publish_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/publish_task.py
@@ -2,10 +2,14 @@ from __future__ import annotations
 
 import os
 import subprocess
+import urllib.parse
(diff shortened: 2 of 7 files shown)


John-P:
I see that uv is included as a dependency of Kraken. However, would a check and error message if uv is not in the `PATH` be useful? Also is a custom `uv_cmd` (as was used with twine previously) still needed? I removed this as I saw uv included in the dependencies and other code shelling out to uv without a variable cmd.

John-P:
> The unused interactive property has been removed. 

Perhaps keeping interactive but with a deprecation warning noting that it does nothing would be more helpful to not break things initially for users.

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -33,31 +33,28 @@ def get_relationships(self) -> Iterable[TaskRelationship]:
...
+            env["UV_PUBLISH_USERNAME"] = credentials[0]
+            env["UV_PUBLISH_PASSWORD"] = credentials[1]
 
-        safe_command = [x.replace(credentials[1], "MASKED") for x in command] if credentials else command
+        safe_command = command
         self.logger.info("$ %s", safe_command)
 
-        returncode = subprocess.call(command, cwd=self.project.directory)
+        returncode = subprocess.call(command, cwd=self.project.directory, env=env)
         return TaskStatus.from_exit_code(safe_command, returncode)
```
John-P:
```suggestion
    try:
        returncode = subprocess.call(command, cwd=self.project.directory, env=env)
    except FileNotFoundError:
        return TaskStatus.failed(
            "The 'uv' command was not found. The python.publish task requires uv to be installed. "
            "See https://github.com/astral-sh/uv for installation instructions."
        )

        return TaskStatus.from_exit_code(safe_command, returncode)
```

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -33,31 +33,28 @@ def get_relationships(self) -> Iterable[TaskRelationship]:
 
     def execute(self) -> TaskStatus:
         credentials = self.index_credentials.get()
```
John-P, line 74: With
`interactive: Property[bool | None] = Property.default(None)`

```suggestion
    def execute(self) -> TaskStatus:
        # Check for the deprecated property
        if self.interactive.get() is not None:
            warnings.warn(
                "The 'interactive' property on the python.publish task is deprecated and has no effect. "
                "uv publish is non-interactive by default in this context.",
                DeprecationWarning,
            )
        credentials = self.index_credentials.get()
```

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -66,12 +63,10 @@ def publish(
     package_index: str,
     distributions: list[Path] | Property[list[Path]],
     skip_existing: bool = False,
-    interactive: bool = True,
     name: str = "python.publish",
     group: str | None = "publish",
     project: Project | None = None,
     after: list[Task] | None = None,
-    twine_version: str = ">=6.0.0,<7.0.0",
```
ulucs, line 74: can we print a warning here to warn the users that this argument has been removed and will be ineffective?

kraken-build/src/kraken/std/python/settings.py
```diff
@@ -90,6 +94,8 @@ def add_package_index(
             If not specified, *alias* must be a known package index (`pypi` or `testpypi`).
         :param upload_url: If the upload url deviates from the registry URL. Otherwise, the upload URL will
             be the same as the *url*,
+        :param check_url: Optional index URL to check for existing files. Otherwise, the index URL will be
+            used.
```
ulucs: Does this behavior come from `uv`? I would have expected to see `index_url` as the default value of `check_url`

John-P: Yes this is from uv and the index_url is the default if check_url is not given to uv. We could explicitly assign check_url to index_url here if that would be more clear.

kraken-build/src/kraken/std/python/tasks/publish_task.py
NiklasRosenstein: Practically nobody is using that argument; according to internal search `kraken-hs` is the only reference for `twine_version` and it pins the minor version of `kraken-build` so this would be safe to remove.

kraken-build/src/kraken/std/python/tasks/publish_task.py
NiklasRosenstein: If we use the `uv` binary that comes with Kraken through it's dependencies, we wouldn't need that check. That UV version is pinned with `.kraken.lock`, which seems like a desirable property; although on the other hand it might mean the `uv` version run by users when invoked manually is likely a different one..

kraken-build/src/kraken/std/python/settings.py
NiklasRosenstein: Do we need the `check_url` configured separately? Is it because the URL cannot be the `index_url` for Artifactory? 

I would consider a `exist_ok` property on the publish task a cleaner way to express the intent.

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -33,31 +33,28 @@ def get_relationships(self) -> Iterable[TaskRelationship]:
...
-            command.append("--non-interactive")
         if self.skip_existing.get():
-            command.append("--skip-existing")
+            command.extend(["--check-url", self.index_check_url.get() or self.index_index_url.get()])
+
+        env = os.environ.copy()
+        if credentials:
+            # See https://docs.astral.sh/uv/guides/package/#publishing-your-package
+            # NOTE: PyPI does not support publishing with username and password anymore,
+            #       should we log a warning if the username is not __token__?
```
NiklasRosenstein: I think that's reasonable! The check could be `"pypi.org/" in self.index_upload_url.get() and credentials[0] != "__token__"`. Would be good to include a bit of info as well like `logger.warning("since 2025-MM-DD, PyPi no longer allows publishing with username/password, see https://...")`

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -33,31 +33,28 @@ def get_relationships(self) -> Iterable[TaskRelationship]:
...
+            env["UV_PUBLISH_USERNAME"] = credentials[0]
+            env["UV_PUBLISH_PASSWORD"] = credentials[1]
 
-        safe_command = [x.replace(credentials[1], "MASKED") for x in command] if credentials else command
+        safe_command = command
         self.logger.info("$ %s", safe_command)
 
-        returncode = subprocess.call(command, cwd=self.project.directory)
+        returncode = subprocess.call(command, cwd=self.project.directory, env=env)
```
NiklasRosenstein: nit: would prefer this style, as it keeps the intent of "we need to pass the full environment not just overrides" more localised

```suggestion
        returncode = subprocess.call(command, cwd=self.project.directory, env={**os.environ, **env})
```

NiklasRosenstein (commented):
Left a few comments; looks good overall! 👍 Not approving just yet due to open discussions

kraken-build/src/kraken/std/python/settings.py
John-P: Sooo...I have looked into this a bit further and twine and uv actually differ in behavior here.

### They determine how to skip files differently

- **Twine** [relies on status codes (409)](https://github.com/pypa/twine/blob/974a8d8b1e63f65ab31936ae39a63adcdf3f47c6/twine/commands/upload.py#L32-L64) which is returned on existing same file name
- **uv** uses content-based hashes (assumes the registry supports SHA hashes in the [simple PyPI API](https://packaging.python.org/en/latest/specifications/simple-repository-api/#simple-repository-json-project-detail)). See the quote below from the uv docs.

### uv `--check_url` with PyPI vs Others

With uv, the `--check-url` option behaves differently for PyPI vs other registries. For PyPI this is not required, because the index URL for PyPI is known to uv (since it is the default registry).

From the [uv documentation](https://docs.astral.sh/uv/guides/package/#publishing-your-package) (emphasis added):

> Even though uv publish retries failed uploads, it can happen that publishing fails in the middle, with some files uploaded and some files still missing. **With PyPI, you can retry the exact same command, existing identical files will be ignored. With other registries, use --check-url <index url> with the index URL** (not the publishing URL) the packages belong to. **When using --index, the index URL is used as check URL**. uv will skip uploading files that are identical to files in the registry, and it will also handle raced parallel uploads. Note that **existing files need to match exactly** with those previously uploaded to the registry, this avoids accidentally publishing source distribution and wheels with different contents for the same version.

So in order for skipping existing files to work, either:
- The index URL must be set, OR
- The check URL must be set (usually the same as index-url). This is in case you want to use a different URL to check files to skip than is used for the regular index API.

### Additional Failures from Different Behaviour

Since uv determine files to skip based on the hash and not the filename, this means that an upload may fail when it would not for twine `--skip-existing`. For example, if the code was changed and an upload is attempted. With twine this would be ok, and files would simply be ignored (unless a new file is added) with no error. However, with uv this would be an error.

### Proposed Action

The index URL is a non-optional attribute `PackageIndex` but was/is unused. Assuming this is correctly set, we could just actually use this which will enable skipping uploading existing files. Then the check URL is not really needed, assuming that having a different check URL is quite uncommon. However, it is not difficult to support this.

Lastly, if you would like the behavior from twine of "I changed my code but didn't bump the version. Do not overwrite the same files and do not error" we would have to check the exit code on upload and ignore it when a flag is set.

NiklasRosenstein: > Note that existing files need to match exactly with those previously uploaded to the registry, this avoids accidentally publishing source distribution and wheels with different contents for the same version.

For our use case (retrying a CI job that failed and partially uploaded some packages) this will likely be insufficient. The archive produced by `uv build` is not deterministic and the content hash of the generated packages will be different on a retry. Hence, it will try to upload _all_ files, and fail again but this time for the files already published.

> The index URL is a non-optional attribute PackageIndex but was/is unused.

When you say unused, you mean in the context of publishing packages? Because it is used, just elsewhere (e.g. to inject the indexex into the `pyproject.toml`). If the index URL is to be the same value as the check URL in practically all cases, I think we can forego a separate check URL.

Or alternatively, we can do both: Add a `check_url` that if set and supported will be used to check for existing packages, but the publish task also gets an `exist_ok` option, which will have the task make sure the same package getting published again is OK and not an error (in which case it will use the `check_url` if available or otherwise `index_url`, or by relying on the behavior of Uv to do that fallback).

But the semantics of an `exist_ok` or `skip_existing` property must be clear. The existing behavior with Twine was to accept conflict responses as you pointed out (409 Conflict), but Uv seems to rely on file hashes. Those are two very distinct semantics that will behave very differently. We might need to implement the check ourselves to recover Twine's `--skip-existing` behavior.

Note that we already implement a similar check for Cargo ourselves: https://github.com/kraken-build/kraken/blob/develop/kraken-build/src/kraken/std/cargo/tasks/cargo_publish_task.py#L153

kraken-build/src/kraken/std/python/tasks/publish_task.py
John-P: env is initialized with os.environ so this would have no effect surely?

kraken-build/src/kraken/std/python/settings.py
John-P: > When you say unused, you mean in the context of publishing packages? 

Yes, sorry if unclear.

Thanks for the clarification on indented behavior here. Looking into it

John-P:
Interesting to see that twine has removed support for skip-existing on non-pypi indexes: https://github.com/pypa/twine/pull/1253

kraken-build/src/kraken/std/python/settings.py
John-P: I have now added a prepare check similar to cargo which will use the index URL and PyPI simple API to check for an uploaded file listing and will skip uploading if the distribution files already exist. This should emulate the behavior of the old twine `--skip-existing`.

I have tested this with the pypiserver and also manually with Artifactory (Which both return the HTML response format).

John-P:
One thing I noticed while working on this is that there is a mix of loguru and logging used here which I was a bit confused about. I am using Task.logger here but also could not get this to output anywhere, so used loguru for testing/debugging. It looks like there is no handler set up? Just wanted to query if I was doing something incorrect there or if this is something that need fixing.

kraken-build/src/kraken/std/python/settings.py
ulucs: Noting here that our python builds are non-deterministic and this difference in behavior will show itself quite often. I am of the opinion that checking the hash and erroring is a desirable approach (even if it differs from what twine does), but I'm deferring to Nik on what Kraken should be doing.

NiklasRosenstein:
Loguru was introduced new a while ago, but changing `Task.logger` to a Loguru logger would be a (potentially very silent) breaking change. I think using `Task.logger` is fine regardless where one has access to it, and otherwise `loguru`.

Did you try increasing the verbosity to get the logs to show?

The intercept handler is configured here:

https://github.com/kraken-build/kraken/blob/7bdab959c52e7fd789813094d2a742ce0423bcea/kraken-build/src/kraken/common/_option_sets.py#L59-L63

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -31,47 +36,166 @@ def get_relationships(self) -> Iterable[TaskRelationship]:
         yield from (TaskRelationship(task, True, False) for task in self.dependencies)
         yield from super().get_relationships()
 
+    def _get_existing_files_from_index(self, project_name: str) -> set[str] | None:
```
NiklasRosenstein: I think this function would do well to live as a standalone function; I don't see a good reason for it to live as a member method. If nothing else, it could be a `@staticmethod` that can then also be tested independently of an instance of the Task (after adding parameters for the index URL and credentials).

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -31,47 +36,166 @@ def get_relationships(self) -> Iterable[TaskRelationship]:
...
+        result = subprocess.run(
+            command,
+            cwd=self.project.directory,
+            env={**os.environ, **env},
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+
+        if all(
```
NiklasRosenstein: Given the manual filtering of the distributions based on the existing files that already exist in the index, wouldn't we expect to never run into this condition?

NiklasRosenstein (commented):
Left two more comments. :)

Very nice job btw., particularly on those tests! 🚀

kraken-build/src/kraken/std/python/tasks/publish_task.py
John-P: Ah yes, this *should* not longer be possible so I think it is now a bit redundant. See https://github.com/kraken-build/kraken/pull/389/commits/b47f973dbdcc3e6c433d993bec8afc18a1f49849

kraken-build/src/kraken/std/python/tasks/publish_task.py
John-P: I have extracted it as a stand-alone function now. See https://github.com/kraken-build/kraken/pull/389/commits/cfa2acf71647aa2fbf112fc9b3e5c54910353ad9

John-P:
Thanks for the review. I think I have addressed the comments on extracting the function and adding some tests for it also. Sorry for the slightly messy conversation log! I did a rebase to try and tidy up the git history after a minor mistake but did not realise it would make the conversation view a mess!

kraken-build/src/kraken/std/python/settings.py
NiklasRosenstein: With the changes added by @John-P to ignore distribution files that are already in the index, I think this can be resolved.

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -31,47 +37,157 @@ def get_relationships(self) -> Iterable[TaskRelationship]:
...
+
+        distributions = self.distributions.get()
+        if not distributions:
+            return TaskStatus.skipped("No distributions to publish.")
+
+        project_name = distributions[0].name.split("-")[0]
+        existing_files = _get_existing_files_from_index(
+            project_name, self.index_index_url.get(), self.index_credentials.get()
+        )
+        logger.debug(f"Existing files: {existing_files}")
```
NiklasRosenstein:
```suggestion
        self.logger.debug("Existing files in index: %s", existing_files)
```

For consistency

kraken-build/src/kraken/std/python/tasks/publish_task.py
```diff
@@ -31,47 +37,157 @@ def get_relationships(self) -> Iterable[TaskRelationship]:
...
+                self.logger.warning(
+                    "Since 2024-01-01, PyPI no longer allows publishing with username and password, "
+                    " see https://blog.pypi.org/posts/2024-01-01-2fa-enforced/"
+                    " and https://docs.astral.sh/uv/guides/package/#publishing-your-package"
+                )
+            env["UV_PUBLISH_USERNAME"] = credentials[0]
+            env["UV_PUBLISH_PASSWORD"] = credentials[1]
+
+        safe_command = command
         self.logger.info("$ %s", safe_command)
```
NiklasRosenstein:
```suggestion
        safe_command = command
        self.logger.info("$ %s", command)
```

Since the command itself does not contain unsafe data

NiklasRosenstein (changes requested):
Almost ready  🚀

NiklasRosenstein (approved):

PR #393, opened 2025-10-09
Bump uv to 0.9

Tpt:


kraken-build/src/kraken/core/system/graph.py
```diff
@@ -356,7 +356,7 @@ def tasks(
         tasks = (self.get_task(addr) for addr in self._digraph)
         if goals:
             # HACK: Cast because of https://github.com/python/typeshed/pull/12472
```
ghadjigeorghiou: We should remove the comment since we are removing the casting

```suggestion
```

kraken-build/src/kraken/core/system/graph.py
```diff
@@ -475,7 +475,7 @@ def ready(self) -> list[Task]:
...
```
ghadjigeorghiou: ditto

```suggestion
```

ghadjigeorghiou (approved):

PR #396, opened 2025-10-17
Allow Mitmproxy 12

Tpt:
First step for Python 3.14 support

### DIFF
--- a/kraken-build/.kraken.py
+++ b/kraken-build/.kraken.py
@@ -1,7 +1,7 @@
 from kraken.std import python
 
 python.install()
-python.mypy(version_spec="==1.16.1")
+python.mypy(version_spec="==1.18.2", python_version="3.12")
 python.ruff(additional_args=["--exclude", "tests/data"])
 python.pytest(
     tests_dir=["src", "tests/kraken_core"],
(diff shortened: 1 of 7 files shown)


kraken-build/.kraken.py
```diff
@@ -1,7 +1,7 @@
 from kraken.std import python
 
 python.install()
-python.mypy(version_spec="==1.16.1")
+python.mypy(version_spec="==1.18.2", python_version="3.12")
```
Ghelfi: Is this change related to mitmproxy bump?

kraken-wrapper/.kraken.py
```diff
@@ -1,7 +1,7 @@
 from kraken.std import python
 
 python.install()
-python.mypy(version_spec="==1.16.1")
+python.mypy(version_spec="==1.18.2", python_version="3.12")
```
Ghelfi: Same here.

kraken-build/.kraken.py
Tpt: Yes, if we don't set python version, mypy uses the `requires-python` field. However, because we are running lints on Python 3.13, uv will install the latest mitmproxy version that requires Python 3.12 and, hence, fail because of type syntax requiring Python 3.12. I have added a comment.

Ghelfi: Thanks for the explanation.

Ghelfi (approved):

PR #397, opened 2025-10-17
Bump default ruff version to 0.14

Tpt:


Ghelfi (approved):
Thanks for the bump. LGTM!

PR #402, opened 2025-10-23
fix: `krakenw` now detects broken build environment when `python -V` fails

NiklasRosenstein:
Python virtual envs break when you remove the Python installation that was used to create the environment.

This seems kind of obvious, but it includes cases where you upgrade your system Python version causing it to live at another path (e.g. `/opt/homebrew/Cellar/python@3.13/3.13.7/bin/python3` and `/opt/homebrew/Cellar/python@3.14/3.14.0/bin/python3`). The Python version is encoded in most kinds of ways Python is installed, especially because the virtual environment's symlinks point to the real path of the Python interpreter that was used to create it and not an intermediate symlink (as that could be updated to point to a newer version of Python, breaking your virtual env in another way).

Recently it's caused issued when we published a new version of the [kraken-base-image](https://github.com/kraken-build/kraken-base-image) where we only pin the minor version of Python, and Uv installs the latest of that minor branch. Our internal CI caches the virtual environment, thus when CI runs with a newer version of the base image with slightly different Python versions, the virtual environments are broken and error instead of being discarded and re-initialized (as is now done with this PR).

The krakenw entrypoint also had a bit of duplicate logic for determining whether a build environment is to be re-used or needs some additional actions (such as calling `uv sync` once more). The environment manager's `install()` method is now called always and it could make the decision to not invoke `uv` if everything seems up-to-date, but since `uv` is so fast I haven't bothered to do that yet.

### DIFF
--- a/.changelog/_unreleased.toml
+++ b/.changelog/_unreleased.toml
@@ -16,3 +16,10 @@ author = "@NiklasRosenstein"
 issues = [
     "https://github.com/kraken-build/kraken/issues/400",
 ]
+
+[[entries]]
+id = "3d3f462c-c28d-41ae-acbf-667ebf78cbca"
+type = "fix"
+description = "`krakenw` now detects broken build environment when `python -V` fails"
+author = "@NiklasRosenstein"
+component = "kraken-wrapper"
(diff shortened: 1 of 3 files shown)


Tpt (approved):
Thank you for finding the issue cause! Makes sense to run `uv sync` everytime

PR #412, opened 2025-11-28
Adds very basic support of Ty

Tpt:
Ty seems to move forward nicely

### DIFF
--- a/kraken-build/src/kraken/std/python/tasks/ty_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/ty_task.py
@@ -4,8 +4,7 @@ import logging
 import os
 from collections.abc import MutableMapping, Sequence
 
-from kraken.common import Supplier
-from kraken.common._fs import intersect_paths
+from kraken.common import Supplier, intersect_paths
 from kraken.core import Project, Property
 from kraken.core.system.aspect import CheckAspect
 from kraken.core.system.task import TaskStatus
--- a/kraken-build/src/kraken/std/python/tasks/ty_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/ty_task.py
@@ -59,7 +59,7 @@ def ty(
     version_spec: str | None = None,
 ) -> TyTask:
     """
-    :param version_spec: If specified, the Mypy tool will be installed as a PEX and does not need to be installed
+    :param version_spec: If specified, the Ty tool will be run via `uv tool run` and does not need to be installed
         into the Python project's virtual env.
     """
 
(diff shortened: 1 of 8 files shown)


kraken-build/src/kraken/std/python/tasks/ty_task.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import annotations
+
+import logging
+import os
+from collections.abc import MutableMapping, Sequence
+
+from kraken.common import Supplier
+from kraken.common._fs import intersect_paths
```
srstevenson: This is now used in the pytest, ruff, and mypy tasks and so is, in practice, a public API. Should we rename the module from `_fs` to `fs`, or re-export it from `__init__.py` in  `kraken.common` as part of the public API?

kraken-build/src/kraken/std/python/tasks/ty_task.py
```diff
@@ -0,0 +1,77 @@
...
+def ty(
+    *,
+    name: str = "python.ty",
+    project: Project | None = None,
+    additional_args: Sequence[str] | Supplier[Sequence[str]] = (),
+    check_tests: bool = True,
+    version_spec: str | None = None,
+) -> TyTask:
+    """
+    :param version_spec: If specified, the Mypy tool will be installed as a PEX and does not need to be installed
```
srstevenson:
```suggestion
    :param version_spec: If specified, the Ty tool will be run via `uv tool run` and does not need to be installed
```

kraken-build/src/kraken/std/python/tasks/ty_task.py
Tpt: Indeed. I have reexported the function to be more in line with the way `kraken.common` is architectured.

kraken-build/src/kraken/std/python/tasks/ty_task.py
```diff
@@ -0,0 +1,76 @@
...
+from collections.abc import MutableMapping, Sequence
+
+from kraken.common import Supplier, intersect_paths
+from kraken.core import Project, Property
+from kraken.core.system.aspect import CheckAspect
+from kraken.core.system.task import TaskStatus
+
+from .base_task import EnvironmentAwareDispatchTask
+
+logger = logging.getLogger(__name__)
```
srstevenson: This is unused.

```suggestion
```

kraken-build/src/kraken/std/python/tasks/ty_task.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import annotations
+
+import logging
```
srstevenson:
```suggestion
```

srstevenson (approved):

PR #415, opened 2025-12-05
Fix publish task failing with "Property is finalized" error

John-P:
The `PublishTask` was failing with:

```text
RuntimeError: Property(PublishTask(...).distributions) is finalized
```

This occurred when `skip_existing=True` because the code attempted to modify the `distributions` property using `.set()` in `prepare()`, but non-output properties are finalized before `prepare()` runs.

## Solution

Marked the `distributions` property as an output property using `Property.output()`. Output properties can be modified after finalization, allowing the `skip_existing` logic in `prepare()` to filter the list using `.set()`.

## Changes

- Changed `distributions: Property[list[Path]]` to `distributions: Property[list[Path]] = Property.output()`
- This allows the property to be modified in `prepare()` when filtering out files that already exist on the package index

### DIFF
--- a/kraken-build/src/kraken/std/python/tasks/publish_task.py
+++ b/kraken-build/src/kraken/std/python/tasks/publish_task.py
@@ -24,7 +24,7 @@ class PublishTask(Task):
     index_upload_url: Property[str]
     index_index_url: Property[str]
     index_credentials: Property[tuple[str, str] | None] = Property.default(None)
-    distributions: Property[list[Path]]
+    distributions: Property[list[Path]] = Property.output()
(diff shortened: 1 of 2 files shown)


kraken-build/src/kraken/std/python/tasks/publish_task.py
NiklasRosenstein: Properties marked as outputs can be mutated after `Task.finalize()`! 🙂 


```suggestion
    distributions: Property[list[Path]] = Property.output()
```

NiklasRosenstein (approved):

PR #427, opened 2026-04-13
Sort Python interpreter candidates for deterministic discovery order

mantoine96:
## Summary

`_get_candidates()` collects Python binaries into a `set[Path]` and iterates over it to yield candidates. Python set iteration order depends on `PYTHONHASHSEED`, which is randomised per-process by default. This causes `find_python_interpreter()` to non-deterministically select different Python versions across runs when multiple versions satisfy the constraint.

## Problem

This matters because `.kraken.lock` is a flat format (no per-version forks). Packages guarded by `python_version` markers (e.g. `backports-tarfile; python_version < "3.12"`) are included or excluded depending on which interpreter was chosen, producing non-reproducible lock files and endless add/remove oscillation in CI-driven lock-file maintenance (observed on multiple repos with 20+ consecutive flip-flop commits).

## Fix

Sort the `set[Path]` by filename before iterating. This ensures deterministic, version-ascending order (`python3.10` < `python3.11` < `python3.12`), so `find_python_interpreter()` always picks the lowest matching version.

This follows what is done for example with uv: https://docs.astral.sh/uv/concepts/python-versions/#discovery-of-python-versions

### DIFF
--- a/kraken-build/src/kraken/common/findpython.py
+++ b/kraken-build/src/kraken/common/findpython.py
@@ -116,7 +116,7 @@ def _get_candidates(
     if path_list is None:
         path_list = os.environ["PATH"].split(os.pathsep)
 
-    commands: set[Path] = set()
+    commands_set: set[Path] = set()
     for path in map(Path, path_list):
         if not path.is_dir():
             continue
@@ -129,7 +129,13 @@ def _get_candidates(
             except PermissionError:
                 continue
             else:
-                commands.add(item)
+                commands_set.add(item)
+
+    # Sort for deterministic iteration order. Without this, set iteration
+    # order depends on PYTHONHASHSEED, which is randomised per-process,
+    # causing find_python_interpreter() to non-deterministically select
+    # different Python versions across runs.
+    commands = sorted(commands_set, key=lambda p: p.name)
 
     # py and python
     for command in commands:
--- a/kraken-build/src/kraken/common/findpython.py
+++ b/kraken-build/src/kraken/common/findpython.py
@@ -181,8 +181,7 @@ def _get_candidates(
 
     if pyenv_versions and pyenv_versions.is_dir():
         pyenv_dirs = [
-            item for item in pyenv_versions.iterdir()
-            if re.match(r"\d+\.\d+\.\d+$", item.name) and item.is_dir()
+            item for item in pyenv_versions.iterdir() if re.match(r"\d+\.\d+\.\d+$", item.name) and item.is_dir()
         ]
         pyenv_dirs.sort(key=lambda p: Version(p.name))
         for item in pyenv_dirs:
--- a/kraken-build/src/kraken/common/findpython.py
+++ b/kraken-build/src/kraken/common/findpython.py
@@ -135,9 +135,12 @@ def _get_candidates(
     # Sort for deterministic iteration order and correct version ordering.
     # Without this, set iteration order depends on PYTHONHASHSEED (randomised
     # per-process), and lexicographic sort places python3.9 after python3.10.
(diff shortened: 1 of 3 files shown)


kraken-build/tests/kraken_common/test_findpython.py
```diff
@@ -0,0 +1,64 @@
...
+    def test_candidates_are_deterministic_across_seeds(self, tmp_path: Path) -> None:
+        """The candidate order must not depend on the hash seed."""
+        bin_dir = tmp_path / "bin"
+        bin_dir.mkdir()
+        for name in ["python3.13", "python3.10", "python3.12", "python3.11"]:
+            _make_executable(bin_dir / name)
+
+        results: list[list[str]] = []
+        for seed in ["0", "1", "42", "99999"]:
+            with patch.dict(os.environ, {"PYTHONHASHSEED": seed}):
```
srstevenson: I don't think this has any effect as `PYTHONHASHSEED` is only read once when the interpreter starts, so the ordering here is deterministic because all the cases are running in the same interpreter process, and I expect this test would pass without your change to `_candidates_from`.

kraken-build/src/kraken/common/findpython.py
```diff
@@ -129,7 +129,13 @@ def _get_candidates(
...
                 continue
             else:
-                commands.add(item)
+                commands_set.add(item)
+
+    # Sort for deterministic iteration order. Without this, set iteration
+    # order depends on PYTHONHASHSEED, which is randomised per-process,
+    # causing find_python_interpreter() to non-deterministically select
+    # different Python versions across runs.
+    commands = sorted(commands_set, key=lambda p: p.name)
```
srstevenson: This is going to sort `python3.10` before `python3.11`, but also before `python3.9`. Do we mind this, or should we sort by version number?

kraken-build/src/kraken/common/findpython.py
srstevenson: Should we sort the pyenv versions we find, too?

mantoine96: Done!

kraken-build/src/kraken/common/findpython.py
mantoine96: Yes we do! I switched to using `packaging.Version` which handles this sort of thing nicely.

kraken-build/tests/kraken_common/test_findpython.py
mantoine96: You're right. This was my mistake. We can't really test for this without running the interpreter multiple times.

kraken-build/src/kraken/common/findpython.py
```diff
@@ -129,7 +130,19 @@ def _get_candidates(
...
+                commands_set.add(item)
+
+    # Sort for deterministic iteration order and correct version ordering.
+    # Without this, set iteration order depends on PYTHONHASHSEED (randomised
+    # per-process), and lexicographic sort places python3.9 after python3.10.
+    # The full path (str(p)) is used as tiebreaker to ensure deterministic
+    # ordering even when multiple binaries share the same name (e.g. several
+    # "python" or "python3" from different PATH directories).
+    def _interpreter_sort_key(p: Path) -> tuple[Version, str]:
+        m = re.match(r"python(\d+(?:\.\d+)*)$", p.name)
```
srstevenson: nit: [`re.compile`](https://docs.python.org/3/library/re.html#re.compile) this outside the loop.

srstevenson (approved):

kraken-build/src/kraken/common/findpython.py
mantoine96: Good point! Fixed :)

PR #434, opened 2026-05-06
Don't install all packages when using uv workspaces

antoinebrl:
Kraken shouldn't enforce that when using uv workspaces all packages must be installed inside the single venv. The concept of workspace in Kraken is quite different than the one in uv.

### DIFF
--- a/kraken-build/src/kraken/std/python/buildsystem/uv.py
+++ b/kraken-build/src/kraken/std/python/buildsystem/uv.py
@@ -350,7 +350,7 @@ class UvManagedEnvironment(ManagedEnvironment):
 
     def install(self, settings: PythonSettings) -> None:
         # --locked is used only when uv.lock is present
-        cmd = ["uv", "sync", "--all-packages"]
+        cmd = ["uv", "sync"]
         if (self.project_directory / "uv.lock").is_file():
             cmd.append("--locked")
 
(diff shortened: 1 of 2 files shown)


kraken-build/src/kraken/std/python/buildsystem/uv.py
```diff
@@ -350,7 +350,7 @@ def get_path(self) -> Path:
 
     def install(self, settings: PythonSettings) -> None:
         # --locked is used only when uv.lock is present
-        cmd = ["uv", "sync", "--all-packages"]
+        cmd = ["uv", "sync"]
```
Ghelfi, line 353: I agree that it shouldn't be enforced. Keeping it a s a possibility would be nice though. Did you consider adding the option (as opt in) through `PythonSettings`?

antoinebrl: I don't see a nice way to inject some optional configuration into the `UvPythonBuildSystem` given it's only instantiated via the `detect_build_system`.

Ghelfi: Hmmm, let's move forward for now and eventually add the flag later on.

Ghelfi (approved):

